PR #47, opened 2017-03-23
improve ko locales

KimDaesap:


nicolasg-playster:
Thanks! The old ones were machine translated...

Next beta will include much better translations support in the UI, so if you're willing to translate these as well it would be awesome.

I'll merge this PR and mark these translations as approved.

NGPixel:
Thanks! The old ones were machine translated...

Next beta will include much better translations support in the UI, so if you're willing to translate these as well it would be awesome.

I'll merge this PR and mark these translations as approved.

KimDaesap:
Awesome wiki project! Thanks!
happy if it can be helpful.
watching this project.

PR #109, opened 2017-05-20
Add table editor

fa93hws:
It give user a hint to generate table from http://www.tablesgenerator.com/ and paste the code back. Further editing in ace editor is allowed if the user want.


fa93hws:
Sorry, the change in package.json is not necessary.
requirejs is not used anywhere.
please ignore that change.

PR #112, opened 2017-05-23
Update Dockerfile

felipeplets:
Add git and OpenSSH to the default Docker image so we can use git over ssh.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
 FROM node:6-alpine
 
+RUN apk update && apk upgrade && \
+    apk add --no-cache bash git openssh
+
 ENV WIKI_JS_DOCKER 1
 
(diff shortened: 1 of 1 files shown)


NGPixel:
Good idea, thanks!

Still a few things to add for proper Docker support but slowly getting there...

felipeplets:
@NGPixel count on me to make it work, I've been working on Docker images and container for some months and may be able to help on it.
 
I had to write my own entrypoint script to make openssh work by default and I'm still facing some issues, I will share it here as I progress on this matter.

Keep the good work!

NGPixel:
That's because the current dockerfile is pointing to wiki.js@latest which is Beta 11.

The upcoming Beta 12 greatly improves support for Docker (and Heroku), both for setup and running the server.

PR #120, opened 2017-05-30
WIP: Usable Dockerfile

Mika56:
Hi,

I've made this Dockerfile for our use at work. It mostly works but needs a few adjustments to be more versatile:

- [ ]  Assumes it is running under a Docker Swarm, and uses `wikijs_config` and `wikijs_privatekey` secrets. Wikijs should probably be updated so it can be configured using environment variables, and not files
- [ ] Builds Git 2.11.2 from source (should probably try to fetch latest Git version), because Wikijs requires Git >= 2.10 (Git config `core.sshCommand` is only available since Git 2.10.0), which is only available in Alpine 3.5+, while `node:7-alpine` uses 3.4 (see nodejs/docker-node#406) (not a bad thing per-see, but could speed-up the compilation process if we were able to use Alpine 3.5)
- [ ] Does not auto-build and upload to Docker Hub. This can be done quite easily through Docker Hub, and only an admin can

Any contribution, idea or remark is more than welcome!

### DIFF
--- a/.docker/entrypoint.sh
+++ b/.docker/entrypoint.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+if [ -f /run/secrets/wikijs_config ]; then
+        cp /run/secrets/wikijs_config config.yml
+fi
+
+if [ -f /run/secrets/wikijs_privatekey ]; then
+        cp /run/secrets/wikijs_privatekey wikijs_privatekey.pem
+        chmod 400 wikijs_privatekey.pem
+fi
+
+exec node server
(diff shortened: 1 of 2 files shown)


NGPixel:
Beta 12 supports environment variables for any config parameter. See [config.heroku.yml](https://github.com/Requarks/wiki/blob/master/npm/configs/config.heroku.yml) for example. The [config.docker.yml](https://github.com/Requarks/wiki/blob/master/npm/configs/config.docker.yml) file should be updated to work the same way.

For compilation + deploy to Docker Hub, I have the project created on Codefresh already. It needs some more config but will soon automatically compile and deploy to Docker Hub for new releases.

PR #121, opened 2017-05-31
change two words from Korean locale, 

sageahn:
IMO,
source : 소스 is better than 출처
move : 이동 is better than 옮김


NGPixel:
@junwonpk Can you review and update the resources in LingoHub if applicable?

NGPixel:
Changed for Beta 12

PR #171, opened 2017-07-20
Fix graceful termination issues

xTNTx:
- replace `process.kill()` which is causing infinite loop of `graceful` callback
- terminate background agent to fix `warn: [AGENT] Lost connection to main server. Exiting...` error


NGPixel:
Thanks!

Issue #176, opened 2017-07-24
Error: Cannot find module 'qs'


theodotos:
I tried to install *wiki.js* with:

```
npm install wiki.js@latest
```

I get a "Cannot find module 'qs'" during the configuration phase.

### Actual behavior

```
✔ Installation succeeded.
? Continue with configuration wizard? Yes
> Browse to http://your-server:3000/ to configure your wiki! (Replaced your-server with the hostname or IP of your server!)
▐   ⠈    ▌ I'll wait until you're done ;)(node:4095) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: Command failed: node wiki configure
/srv/wiki.js/node_modules/yargs/yargs.js:1079
      else throw err
           ^

Error: Cannot find module 'qs'
    at Function.Module._resolveFilename (module.js:485:15)
    at Function.Module._load (module.js:437:25)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/srv/wiki.js/node_modules/express/lib/middleware/query.js:16:10)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/srv/wiki.js/node_modules/express/lib/application.js:20:13)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)


(node:4095) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
▐ ⠂      ▌ I'll wait until you're done ;)
```

### Expected behavior

*wiki.js* waiting for the configuration to finish

### Steps to reproduce the behavior

I am using Ubuntu 16.04.2. My steps up to now:

#### Install *curl*, *Node.js v8.x* and *build-essential*:

```
apt -y install curl
curl -sL https://deb.nodesource.com/setup_8.x | bash -
apt -y install nodejs build-essential
```

#### Install *MongoDB v3.4*

```
apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 0C49F3730359A14518585931BC711F9BA15703C6
[...]

PR #180, opened 2017-07-25
fix: Added missing module 'qs'

bennyn:
Resolves Requarks/wiki#176


NGPixel:
This issue is now fixed. But adding the `qs` package wouldn't actually fix the root problem, as `qs` is not a direct dependency of this project, so the npm dependency install wasn't working like it should. All dependencies are now prebuilt so these issues shouldn't occur anymore.

PR #192, opened 2017-08-04
server locales properly applying

### LINKED ISSUE #191: Server Locales not applying to Client i18n
### Actual behavior
Editing `server/locales/$lang/browser.json` won't apply to `assets/js/i18n/$lang.json` even after build.
`assets/js/i18n/$lang.json` is identical to `server/locales/en/browser.json`

### Expected behavior
`server/locales/en/browser.json` should only be used at missing values of `server/locales/$lang/browser.json`, since `en` is the default.

just go check `assets/js/i18n/$lang.json` everything would be in english.

### Solution
I found that `.build/_tasks.js` was reading locale file from wrong directory!

DevSusu:
fix #191


NGPixel:
👍

PR #249, opened 2017-10-01
function enhancement: multiple directory support

time-river:
This is the function enhancement.
Original, the `host` must be domain in `config.yml` file,  sucn as `example.com`.
Now, I modify some source code, make `example.com/wiki` could also be used.(Attemtion, `/` can't appear in the end.)
I have tested manually, It works well.

Here is an example showing how to use this feature:
```
# config.yml
title: example
host: '//example/wiki'
port: 8000
```
```
# nginx config, too bad, but it could work.
    location ~ ^/fonts/(nucleo-outline.woff2?|wiki-js.woff2?)$ {
        root /home/wiki/wiki.js/assets;
    }
    location /js/ace/ace.js {
        root /home/wiki/wiki.js/assets;
    }
    location /js/simplemde/simplemde.min.js {
        root /home/wiki/wiki.js/assets;
    }

    location /socket.io {
        proxy_pass http://localhost:8000/;
        include proxy_params;
        proxy_redirect off;
    } 

    location /wiki/ {
        client_max_body_size 10m;
        include proxy_params;
        proxy_pass http://localhost:8000/;
        proxy_redirect off;
    }
```

### DIFF
--- a/server/views/auth/login.pug
+++ b/server/views/auth/login.pug
@@ -6,15 +6,15 @@ html(data-logic='login')
     meta(name='viewport', content='width=device-width, initial-scale=1')
     meta(name='theme-color', content='#009688')
     meta(name='msapplication-TileColor', content='#009688')
-    meta(name='msapplication-TileImage', content='/favicons/ms-icon-144x144.png')
+    meta(name='msapplication-TileImage', content=appconfig.host + '/favicons/ms-icon-144x144.png')
     title= appconfig.title
 
     // Favicon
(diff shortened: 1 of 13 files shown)


time-river:
There are still some regret although it could work well.
If the nginx configure file looks like this:
```
    location /wiki/ {
        client_max_body_size 10m;
        include proxy_params;
        proxy_pass http://localhost:8000/;
        proxy_redirect off;
    }
```
It could be better.

PR #312, opened 2017-11-10
Fix port checking

crackcomm:
I would not expect a negative, just a `null` - hopefully, even though it cannot be achieved without changing `app/data.yml` so it is never actually reached in most use cases and `process.env.PORT` cannot be used

I would advise removing default port from `app/data.yml` to fix that.

### DIFF
--- a/server/libs/config.js
+++ b/server/libs/config.js
@@ -42,7 +42,7 @@ module.exports = (confPaths) => {
 
   // Check port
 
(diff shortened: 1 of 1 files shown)


NGPixel:
It's not negative port checking, it checks for 0 (`appconfig.port < 1` and not `appconfig.port < 0`).

PR #320, opened 2017-11-17
change to relative addressing

### LINKED ISSUE #296: Problem with host and port configuration
### Actual behavior
After installation I have configured wiki to start at 3000 port, but it starts the first page there only. All other links were configured to localhost:80 and there was a problem with fetching styles and non-default pages.
### Expected behavior
I expect to have all pages and styles configured to port I wrote during configuration.
1. Install wiki on ubuntu linux 16.04.
2. configure it to use port 3000
### Possible fix
I fixed this problem with hardcoding `https://localhost:3000` in `config.yml` and restarting the server.
I suppose there is some problem with host and port concat.

robert-danowski:
this pull request resolve #296 and allow access from proxy server, multidomains etc.

### DIFF
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -107,7 +107,7 @@ i18next
   .use(i18nextXHR)
   .init({
     backend: {
-      loadPath: siteRoot + '/js/i18n/{{lng}}.json'
+      loadPath: '/js/i18n/{{lng}}.json'
     },
     lng: siteLang,
     fallbackLng: siteLang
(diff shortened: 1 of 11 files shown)


NGPixel:
These changes would completely break installations in sub-directories. You can already use proxy servers without any issues in the current version...

2.0 will have proper support for multidomains.

robert-danowski:
You have right that break installations in sub-directories, but... for example if you set http://example.com:3000 as a host, there will be problem if you set reverse proxy at https://example.com because in code there are addresses to http://example.com and your browser inform you about CORS problems (Cross-Origin Resource Sharing), the multidomain and proxy is much more complicated...

NGPixel:
No, you simply need to set the `host` value in config.yml to `https://example.com`.

PR #356, opened 2017-12-19
Strip out edit and create from start of path in rights

### LINKED ISSUE #147: Giving write permissions to subpage requires granting write permissions to /edit/... and /create/... page
### Actual behavior
I just wanted gave one of my users write permissions to a subpage of my wiki. The wiki showed the `Edit` and `Create` button as expected, however, when trying to edit or create a page I got the `Permission denied` page. This kind of makes sense because I only granted read and write permissions to all paths that start `/subpage` and creating a pages opens `/create/subpage` which technically speaking is not allowed.

### Expected behavior
Even though this technically makes sense, I find it counter-intuitive to allow editing a page, but refuse access to the editing page of that page. A solution could be to add the `edit` or `create` to the end of the URL instead of to the beginning. F.e. `wiki.com/subpage/edit` instead of `wiki.com/edit/subpage`. So, when granting write permission to all paths that start with `/subpage`, the edit and create pages are automatically included.
It also not a major issue. But it is a bit strange.

Grant user write permissions to subpage and try editing that page as that user.

amaanm:
Strips the edit and create url segments from the start of the path, which fixes the issue that requires you to add extra permission sets for `/create/...` and `/edit/...`

Fixes #147

### DIFF
--- a/server/libs/rights.js
+++ b/server/libs/rights.js
@@ -56,7 +56,7 @@ module.exports = {
       manage: false
     }
     let rt = []
-    let p = _.chain(req.originalUrl).toLower().trim().value()
+    let p = _.chain(req.originalUrl).toLower().trim().replace(/^\/(edit|create)/, '').value()
 
     // Load user rights
 


NGPixel:
👍

PR #381, opened 2018-01-24
Propose to split out 2.1 milestone

johnnyshields:
Currently there is a long queue of development items in the 2.0 milestone. Because master is unstable and under heavy changes, currently no one other than @NGPixel can contribute to open source development.

This is not to disparage @NGPixel's great work, but open source works best when an entire community can contribute. In this light I would like to suggest that we do all tickets which are expected to cause heavy changes first, then do "cosmetic" or "add-on" type features which can be implemented without major changes to the core in the 2.1 milestone while allowing 2.0 to stabilize. This will be hopefully allow the community to contribute some of the 2.1 patches so we all get what we want sooner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Current and upcoming milestones *(major features only, see the [changelog](https
   - MongoDB is being phased out in favor of PostgreSQL + Redis. An upgrade tool will be provided to migrate existing data to the new system.
   - Node.js 8.9 LTS or later is now the minimum supported version.
 
-**2.0 Milestones**:
+**2.0 Milestone**:
 - [x] GraphQL API
 - [x] Migrate to PostgreSQL + Redis datastore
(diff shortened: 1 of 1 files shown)


NGPixel:
I agree, but the goal was never to have 100% of the proposed features in 2.0 before having a single release. There will be an alpha release as soon as the new renderer and article view is working, by next month hopefully.

I've moved non-critical features in a 2.1 milestone (not exactly what you listed as editor and search are critical features, which you proposed for 2.1).

2.0 is being rebuilt from the ground up to be fully modular and to address errors / oversights of the 1.x architecture. This means it's not possible to have a stable branch straight from the beginning, as it's not using the 1.x code as its base.

But like I said, once the new renderer and page view is in a working state, a release will be created and it'll be possible for other devs to contribute at this point.

I'll close this PR for now, but feel free to comment.

johnnyshields:
@NGPixel I appreciate your thoughtful response. You da man!!!

PR #462, opened 2018-03-19
Add autoRegisterIds for Slack unregistered accounts registration

cchan:
[first of all, huge compliments on the onboarding process and the wiki itself - it's a beautiful and intuitive experience :) ]

Please tell me what things I might need to fix! First time contributing, our team needed all slack users to be able to auto-register.

Thanks! :)

[one thing - I'm unsure about whether providerId is being set correctly in the case of Slack]

### DIFF
--- a/config.sample.yml
+++ b/config.sample.yml
@@ -80,6 +80,7 @@ auth:
     clientSecret: GITHUB_CLIENT_SECRET
   slack:
     enabled: false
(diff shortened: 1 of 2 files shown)


cchan:
Similar to #302 but allows only Slack registrations from a limited set of team IDs. Too similar?

PR #464, opened 2018-03-20
turn off rgba on thumbnails to avoid being unable to upload files whi…

Gnurdle:
when you try to upload an image that when run through .contain() results in transparency being created, then converting the thumbnail, and thus the upload fails.


Gnurdle:
fixes uploads that result in transparency in the thumbnail == failed upload

Gnurdle:
goober

example of un-uploadable file w/o this fix

NGPixel:
👍

PR #469, opened 2018-03-22
Add autoRegisterIds for Slack unregistered accounts registration

cchan:
[first of all, huge compliments on the onboarding process and the wiki itself - it's a beautiful and intuitive experience :) ]

Please tell me what things I might need to fix! First time contributing, our team needed all slack users to be able to auto-register.

Thanks! :)

[one thing - I'm unsure about whether providerId is being set correctly in the case of Slack]

This may be similar to #302.

### DIFF
--- a/config.sample.yml
+++ b/config.sample.yml
@@ -80,6 +80,7 @@ auth:
     clientSecret: GITHUB_CLIENT_SECRET
   slack:
     enabled: false
+    allowDomains: []
     clientId: 'SLACK_CLIENT_ID'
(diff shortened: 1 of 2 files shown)


cchan:
@NGPixel thoughts?

NGPixel:
v1 doesn't have proper support for default access rights for social connectors (even the current implementation for LDAP and Azure AD is a bit hacky). While your solution might work, it's a very specific fix for your scenario, and I'd like to avoid adding more workarounds for each different provider.

This is already fixed in v2. So keep using that fix in your installation for v1 for now.

PR #605, opened 2018-08-14
Added listen_address configuration option

chathaway-codes:
Hi there!
I had a need to restrict access to localhost, and couldn't find a way to do that in the documentation. This change allows setting listen_address in config.yml, the setting defaults to 0.0.0.0 (which binds all addresses, AFAIK).

### DIFF
--- a/server/app/data.yml
+++ b/server/app/data.yml
@@ -7,6 +7,7 @@ defaults:
   config:
     title: Wiki
     host: http://localhost
+    listenAddress: 0.0.0.0
     port: 80
     paths:
       repo: ./repo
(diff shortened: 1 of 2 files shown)


server/app/data.yml
```diff
@@ -7,6 +7,7 @@ defaults:
   config:
     title: Wiki
     host: http://localhost
+    listen_address: 0.0.0.0
```
NGPixel: use camelCase as this is the standard for this project

chathaway-codes: Fixed

PR #608, opened 2018-08-16
feat: add allowed domain for google auth

the-cybersapien:
Hi,

Google auth has support for including only verified domains. 
This minor change in the config and passport auth provider makes use of that support in two steps:
1. It appends the `hd=YOUR_DOMAIN` query parameter for auth redirect.
> The first part, although provides a good UI from Google to include the allowed domain on the sign in page, it is not entirely functional as Google will still allow a different email to pass.
2. Then comes part 2, in `server/libs/auth.js` we check the result of the auth callback and verify that the email is indeed from a valid domain. 

For this PR, I have only added support for just 1 allowed domains, although I have a feeling this may work for multiple domains as well and would love to test it out based on your feedback.

### DIFF
--- a/config.sample.yml
+++ b/config.sample.yml
@@ -66,6 +66,7 @@ auth:
     enabled: true
     clientId: GOOGLE_CLIENT_ID
     clientSecret: GOOGLE_CLIENT_SECRET
+    hd: 'example.com' # This is optional and can be removed
   microsoft:
     enabled: true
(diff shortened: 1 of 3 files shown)


the-cybersapien:
@NGPixel would really like to hear your thoughts here. This one feature is stopping my organization for using this wiki as an organization-wide solution. 😄

PR #621, opened 2018-09-01
fix: console logger module path

winkelement:


NGPixel:
Fixed in 5919d14670ee028fffa0419b1b2a57ba8fdee7bd

PR #635, opened 2018-09-17
Remove non-existing mdi-dependency

baer95:
The npm package `mdi` does not exist under this name anymore, which leads to `yarn install` failing.

This is just a minor error, but it makes starting the dev build hard for people that want to look into the project and contribute.

### DIFF
--- a/client/components/admin/admin-system.vue
+++ b/client/components/admin/admin-system.vue
@@ -105,17 +105,10 @@
 <script>
 import _ from 'lodash'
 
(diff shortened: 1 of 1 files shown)


NGPixel:
`mdi` is not an npm package, it's an alias defined in webpack for `vue-material-design-icons`. The issue was the recent change to start case for filenames (`mdi/Cube` instead of `mdi/cube`). It's now fixed in the dev branch.

PR #640, opened 2018-09-21
Added new entries and corrected existing ones in Russian localization

Vispiano:
Added new entries and corrected existing ones in Russian localization


NGPixel:
👍 Thanks

PR #704, opened 2018-11-23
Fix status code on page not found

ratzxx:


ratzxx:
Without this commit, wiki returns 304 or 200 code when page not found, this is bad for SEO.

PR #735, opened 2019-01-03
fix: unescape ampersand for TeX math

hoegh:
Unescaped ampersands are important for TeX based math, but the HTML rendered from Markdown converts & to the HTML entity form &amp;

E.g. matrixes are written like this:

```
$$
 \begin{matrix}
  a & b & c \\
  d & e & f \\
  g & h & i
 \end{matrix}
$$
```

Alignment is done like this:

```
$$
\begin{align*}
e^x &= 1 + x + \frac{x^2}{2} + \frac{x^3}{6} + \cdots \\
    &= \sum_{n\geq 0} \frac{x^n}{n!}
\end{align*}
$$
```

There really is no client-side work-around for this. In the best case extra "amp;" will be inserted.

The fix intercepts the TeX formatted snippet just before it is passed to MathJax for rendering, and unescapes ampersands. MathML should be left unaffected.

The fix also unescapes < and > for convenience, though there is the alternatives `\lt` and `\gt` in TeX, that could be used as workarounds. This allows for expressions like:

`$$ a>b \iff b< a $$`

As long as MathJax renders the result to SVG there should be no security issues connected with unescaping. And MathJax should handle the necessary escaping if any other form is rendered.

It should even be safe to do literal text in formulas, like the following sillyness:

`$$ a=2 \text{ iff $n>0$ } $$`


NGPixel:
Kept for future reference in 2.0 as #883

PR #744, opened 2019-01-19
misc: Match sample git URL to auth type

mdmower:
When stepping through graphical configuration, the sample git URL is
https while the sample type is ssh. This leads to confusion, as it
appears that the https URL should be inserted into the box even when ssh
authentication is used. While the input placeholder could clear up this
confusion, it is hidden by the default value. Change the default value
to match the auth type.


NGPixel:
Closing as no longer relevant for 2.x

PR #821, opened 2019-04-27
fix: error messages are not shown in setup

ayZagen:
vee validate scopes are used but they are missing in errors parameter

### DIFF
--- a/dev/webpack/webpack.dev.js
+++ b/dev/webpack/webpack.dev.js
@@ -35,6 +35,7 @@ module.exports = {
     pathinfo: true,
     crossOriginLoading: 'use-credentials'
   },
+  devtool: 'cheap-module-eval-source-map',
   module: {
     rules: [
       {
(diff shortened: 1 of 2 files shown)


dev/webpack/webpack.dev.js
```diff
@@ -35,6 +35,7 @@ module.exports = {
     pathinfo: true,
     crossOriginLoading: 'use-credentials'
   },
+  devtool: 'cheap-module-eval-source-map',
```
NGPixel: Submit a different PR for this and make it optional, as it slows down the build.

karolherbst: maybe just merge the commit without that change? 21 days is a bit long for such a trivial change and I was running into this issue as well and though the setup just wasn't working at all...

ayZagen:
sorry, I was on a vacation. I will resolve this as soon as I return back to my computer :)

PR #880, opened 2019-06-15
fix: locale override not loading

comran:
Locale override options were not working for me with the latest code due to fs.readFileAsync no longer existing in the latest node version. Error was silence too, so there was no feedback to the developer about the issue occurring. This patch should fix it though


NGPixel:
The fix is actually much simplier. We're using node-fs-extra which already promisifies built-in methods. They removed the Async suffix in latest versions, so that's the only change needed. Also we don't want to throw the error as that would break the application if you're not actually testing locales.

I edited your PR to reflect this.

Issue #900, opened 2019-07-02
JWT allows inactive users to use wikijs


fourgates:
**Describe the bug**
Users are still able to access permission-based content on wikijs after their `isActive` flag in the db has been set to false. This will occur if the user has a valid JWT. 

**To Reproduce**
1. Log in to wikijs
2. Set the logged in users `isActive` flag to false
3. User should still be able to view wikijs content that requires permission

**Expected behavior**
If a user is not active they should be treated as a guest. It might make sense to ensure the JWT user is active on each request. 

**Host Info (please complete the following information):**
 - OS: Windows
 - Wiki.js version: 2.0.0-beta.180
 - Database engine: MSSQL

regevbr:
@NGPixel you can close this issue as it is already fixed

wis-ekubesh:
@regevbr do you have any information when it will be available to fetch and use?

NGPixel:
@wis-ekubesh You can already achieve that by setting the token expiration to 1s under Admin > Authentication. If you ever need to ensure a user is logged out, you can invalidate the signing keys from Admin > Utilities.

The default settings are a balance of performance and security. If they don't work for you, you can change them.

wis-ekubesh:
Token expiration 1s it's solution, but as you said before - it will have performance impact. 

I think solution could be tokens storage, for example - redis, where you will be able to revoke token for user.

For example if I deactivated user - all his token can be revoked  and he will be automatically logged out.
And in case when user want to terminate all his active sessions - it will be easy to implement. 

Don't you find it useful?

regevbr:
@wis-ekubesh not sure when it was enabled but you should know that JWT comes with a cost. That cost is that you can't invalidate a token immediately. 

The solution you are proposing effectively renders JWT non relevant as it also tries to solve the massive user validation costs on each request.

@NGPixel 's suggestion to reduce the token expiration will effectively achieve what you want, but as he said, with a cost.

About your issues:

1. This is indeed still an issue - I will open a new issue and fix it soon.
1. This does not happen. When the token expires, a token refresh goes on and it creates a new token with the latest permissions. If you can reproduce the issue, please open a new issue and I will fix it. You can try the following flow to see if it is a real issue:
    1. Create a user and give him some permissions
    1. Set the token expiration to something small, let's say 1 minute
    1. Log in as that user
    1. change the permissions of the user
    1. Wait until the token expires and see if the user can still use his removed permissions
[...]

PR #901, opened 2019-07-03
issue 900 - fixed bug with allowing inactive users with a valid jwt t…

fourgates:
https://github.com/Requarks/wiki/issues/900 - fixed bug with allowing inactive users with a valid jwt to continue viewing non-guest content. It has been documented in the wikijs issue. 

This is my first Wiki.js PR and I have never used this stack before, so let me know if I am not following conventions.  I used `authentication\local\authentication.js` as a reference on how to load a user.

### DIFF
--- a/server/core/auth.js
+++ b/server/core/auth.js
@@ -125,7 +125,16 @@ module.exports = {
           return next()
         }
       }
-
+      // JWT is valid but user is NOT active
+      if (user) {
+        const dbUser = await WIKI.models.users.query().findOne({
+          email: user.email,
+          providerKey: 'local'
+        })
+        if (dbUser && !dbUser.isActive) {
+          user = undefined // user will become a guest
+        }
+      }
       // JWT is NOT valid, set as guest
       if (!user) {
         if (WIKI.auth.guest.cacheExpiration.isSameOrBefore(moment.utc())) {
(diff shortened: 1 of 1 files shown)


fourgates:
https://github.com/Requarks/wiki/issues/900

server/core/auth.js
```diff
@@ -125,7 +125,16 @@ module.exports = {
           return next()
         }
       }
-
+      // JWT is valid but user is NOT active
+      if (user) {
+        const dbUser = await WIKI.models.users.query().findOne({
+          email: user.email,
+          providerKey: 'local'
```
NGPixel: Why is local hardcoded? Should be using whatever provider the user currently has.

fourgates: I updated the PR to use the provider on the user.

NGPixel:
Just noticed that this queries the DB on every request, which is a big no go. The check should ideally be done in the `refreshToken()` method of the users model. I'll make the necessary changes (unless you want to it).

fourgates:
@NGPixel that looks like it would be an easy change. Besides the obvious of not hitting the db on every request, why do you think thats the best place? When does it get called?

fourgates:
@NGPixel did you address this issue?

regevbr:
@NGPixel you can close this pr. The fix suggested here is indeed not good and validation on token refresh is already implemented

PR #921, opened 2019-07-15
fix printing styles

heunghingwan:
when using browser print out pages, menu panel and header is not hiding, this PR fix it.


NGPixel:
Your css classes were a bit too general and could hide unexpected elements. I made the necessary changes in 18b05314e9a4a7f03f3599636d2662c9a3f543ea to properly display the page in print mode.

PR #939, opened 2019-07-25
Feature/menu and search result permissions

fourgates:
This is far from perfect, but I thought it could be a starting point. This is working for my specific use cases. @NGPixel I was hoping you could give me some direction on how I can improve this so that it can make its way into one of the next releases. It seems like a future feature would be to have a menu item mapped to a page, but that doesn't seem to be working currently. When I select `Page` as a target type the `Select Page` button doesn't seem to do anything. 

Therefore, to get the nav to work I used an `External Link` `target type` with the target being the path for the page. e.g. if I have `Page B` with a path of `page-b` the target is set to `/page-b`.

### DIFF
--- a/server/modules/search/db/engine.js
+++ b/server/modules/search/db/engine.js
@@ -19,7 +19,7 @@ module.exports = {
    * @param {String} q Query
    * @param {Object} opts Additional options
    */
-  async query(q, opts, user) {
+  async query(q, opts) {
     let results = await WIKI.models.pages.query()
       .column('id', 'title', 'description', 'path', 'localeCode as locale')
       .where(builder => {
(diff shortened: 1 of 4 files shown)


server/modules/search/db/engine.js
```diff
@@ -52,6 +52,7 @@ module.exports = {
         })
       })
       .limit(WIKI.config.search.maxHits)
+    results = results.filter(page => WIKI.auth.checkAccess(user, ['read:pages'], page))
```
NGPixel: Move the access check outside of the search engine module, into the page resolver instead. Otherwise this logic would be duplicated in every search module which is not necessary.

fourgates: sorry im late on this, where / what is the page resolver?

fourgates:
looking at this conflict, i think maybe this bug has already been resolved?

server/modules/search/db/engine.js
dlbone: Wouldn't this be good for searching through the wiki and hiding results that the user doesn't have permission for?

PR #1004, opened 2019-08-30
Add keycloak auth provider

D4uS1:
Added an authentication provider using the npm package @exlinc/keycloak-passport.
This makes it possible to authenticate against the OpenID protocol of keycloak.
The general OIDC implementations where not able to obtain the access token and user information for some reason. 
Don't know what to set as Author and website, hence i choose github username and fork url.


NGPixel:
Thanks! :+1:

Issue #568, opened 2018-07-07
[FEAT] Storage - AWS S3 Module


NGPixel:
Create AWS S3 storage module

hk1997:
Can you explain the issue further, what we will be storing in s3?

andrewsim:
i can help implementing this feature (and also #567, since they both share the same S3 API)

PR #1015, opened 2019-09-06
Provide basic implementation of AWS S3 and Digitalocean Spaces storage module

andrewsim:
Provide the most basic of implementation for AWS S3 storage module - see #568 and the guideline provided in https://docs.requarks.io/dev/storage.

Creating this PR to get the conversation started.


server/modules/storage/s3/definition.yml
```diff
@@ -1,11 +1,32 @@
 key: s3
 title: Amazon S3
 description: Amazon S3 is a cloud computing web service offered by Amazon Web Services which provides object storage.
-author: requarks.io
+author: andrewsim
 logo: https://static.requarks.io/logo/aws-s3.svg
 website: https://aws.amazon.com/s3/
+isAvailable: true
+supportedModes:
+  - push
```
andrewsim: Clarification or documentation needed for `push`, `pull` and `sync` mode. I have only added these here to get rid of errors on the front-end.

server/modules/storage/s3/storage.js
```diff
@@ -1,23 +1,76 @@
...
+
+/**
+ * AWS S3 Client
+ */
+let s3 = null
+
+/**
+ * Get file extension based on content type
+ */
+const getFileExtension = contentType => {
```
andrewsim: i notice that this gets duplicated in other places, i.e., in the `git` and `disk` storage modules. any plans to refactor them?

NGPixel: Indeed, I'll make it available as part of the page object.

server/modules/storage/s3/definition.yml
NGPixel: **Bi-directional (sync)**
In bi-directional mode, content is first pulled from the storage target. Any newer content overwrites local content. New content since last sync is then pushed to the storage target, overwriting any content on target if present.

**Push to target**
Content is always pushed to the storage target, overwriting any existing content. This is safest choice for backup scenarios.

**Pull from target**
Content is always pulled from the storage target, overwriting any local content which already exists. This choice is usually reserved for single-use content import. Caution with this option as any local content will always be overwritten! (edited) 

Only push must be implemented. Pull and sync are optional. They are mostly for the git module which can easily provide these capabilities.

server/modules/storage/s3/storage.js
andrewsim: let me do that as part of this PR, unless you have already done so?

andrewsim:
@NGPixel I have added the following commits, kindly review:

* 4a6a6398be60b90ba0fcf998d4591eaa2de3f440 Refactored S3 compatible storage module logic into `/modules/storage/s3/common.js`. If we encounter another S3 compatible request, we only need the following four storage module properties and the implementation will be straight forward:
  * `endpoint`
  * `bucket`
  * `accessKeyId`
  * `secretAccessKey`

  If we ever decided to add **Bi-directional** and/or **Pull**, we only need it do it once.

* b0c39dc0d73301cac8c88db46559e912e773e053 I have refactored `getFileExtension()` into the `page` object. I feel the logic to get file path (see [this line](https://github.com/Requarks/wiki/pull/1015/commits/b0c39dc0d73301cac8c88db46559e912e773e053#diff-00c29ad0db170ec8a7ebf4207a009f89R8)) can also be abstracted. Looking into the remaining storage modules, almost all of them are file based, it only means that it will be used very often in the storage module.

* a84e89f56228b205285eb7700082a0f35b3d3baa This shows how we can support other S3 compatible storage.

server/modules/storage/s3/definition.yml
andrewsim: thanks for the clarification, i will leave **pull** and **sync** out for now. do give me a shout if it is needed in the future and i'm happy to open another pull request.

server/modules/storage/s3/common.js
```diff
@@ -0,0 +1,64 @@
+const S3 = require('aws-sdk/clients/s3')
+
+/* global WIKI */
+
+/**
+ * Deduce the file path given the `page` object and the object's key to the page's path.
+ */
+const getFilePath = (page, pathKey) => {
+  const fileName = `${page[pathKey]}.${page.getFileExtension(page.contentType)}`
```
NGPixel: `getFileExtension()` shouldn't have any argument.

andrewsim: goot spot! addressed in 8861944727cf15d5022f10aa450f0f6385afcd5d

Issue #567, opened 2018-07-07
[FEAT] Storage - DigitalOcean Spaces Module


NGPixel:
Create a DigitalOcean Spaces storage module

DO Spaces is technically compatible with S3 API interface. Look into possibly using the same logic as the AWS S3 storage module.

PR #1016, opened 2019-09-06
[WIP] Provide basic implementation of Digitalocean Spaces storage module

andrewsim:
Basic implementation of #567, almost a copy and paste from #1015, creating this PR to get conversation started and to show how similar both these PRs are. The only difference are:

* What we call each properties in `definition.yml` (`bucket` vs `space`)
* How we constract the S3 client (`region` vs `endpoint` override)

We should give some thoughts about how we can abstract S3 compatible storage, once that is done, implementation of other S3 compatible storage will be a breeze.

### DIFF
--- a/server/modules/storage/digitalocean/definition.yml
+++ b/server/modules/storage/digitalocean/definition.yml
@@ -4,7 +4,7 @@ description: DigitalOcean provides developers and businesses a reliable, easy-to
 author: requarks.io
 logo: https://static.requarks.io/logo/digitalocean.svg
 website: https://www.digitalocean.com/products/spaces/
-isAvailable: false
+isAvailable: true
 supportedModes:
   - push
 defaultMode: push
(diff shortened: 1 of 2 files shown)


server/modules/storage/digitalocean/definition.yml
```diff
@@ -16,7 +16,6 @@ props:
     hint: The DigitalOcean datacenter region where the Space will be created.
     default: nyc3
     enum:
-      - ams3
```
andrewsim: Not sure if `ams3` should be removed until `ams3` is fully available again, currently you can't create new space in this region and existing space will be rate limited: https://www.digitalocean.com/docs/platform/release-notes/2019/spaces-ams3/

NGPixel: I would leave it in, as this is a temporary issue and some users might already be using a space in this region.

NGPixel:
A common library could be created in the `s3` module folder, which could then be imported in both the `s3` module (`require('./common')`) and the `digitalocean` module (`require('../s3/common')`).

andrewsim:
closing this, implementation is provided in #1015

PR #1066, opened 2019-09-27
Plantuml markdown

ethanmdavidson:
uses [markdown-it-plantuml](https://www.npmjs.com/package/markdown-it-plantuml).
Doesn't work in preview pane yet. Does the preview pane use a different rendering pipeline?


NGPixel:
Yes it's a different pipeline for now, as it's running client-side. I'll add a note to add this module to it as well.

PR #1084, opened 2019-10-04
Fix go to headers for html-core renderer

hieptuanle:
Really like the table content on client view and the feature to jump to a section by clicking the header. Unfortunately this feature is missing for renderer `html-core` I did a quick update to restore this feature.


NGPixel:
Closing this PR as it's been resolved in 37ce116dd2bee0471650c335ffb3ccf84323e5b8

Your implementation would not support non-latin locales and did not add the anchor element inside the header.

The markdown anchor extension was removed in favor of using this global implementation across all editors.

PR #1111, opened 2019-10-18
Specify navigation visibility base on user's groups

hieptuanle:
I think it would be nice if we can choose which group can see which navigation item, since there are already settings to make some pages be visible from certain user groups only.

Please let me know if I need to change something.


davekopecek:
Yes Please!  I would love to see this implemented.

JBauto:
Any update on this feature? Thanks!

dprensha:
Would love this as well!!!

PR #1118, opened 2019-10-21
Actually use path variables as given in default config

hashworks:
According to the [defaults](https://github.com/Requarks/wiki/blob/master/server/app/data.yml#L60) one can set the data and content paths in the config file. However that wasn't used anywhere until now (besides once in `purge-uploads.js`).

I think the content path isn't used anywhere? Should I remove that one?

### DIFF
--- a/config.sample.yml
+++ b/config.sample.yml
@@ -105,3 +105,11 @@ uploads:
 # download the offline files for sideloading.
 
 offline: false
+
+# ---------------------------------------------------------------------
+# Paths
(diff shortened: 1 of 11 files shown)


NGPixel:
Both are remnants from v1. They are no longer used.
However being able to set a custom data path is a good idea.

I would change the default config to just `dataPath` instead of `data.path`.

Also avoid populating the global WIKI object with `paths`.
Simply use the `WIKI.config.dataPath` when needed.

PR #1237, opened 2019-11-19
Openshift/Kubernetes integration changes

canokcelik:
This merge covers Openshift/Kubernetes **configmap compatibility** for **config.yml** by changing its location to config directory. In this way, it's simple to mount config.yml as configmap.

In addition, the change includes **sideload** directory into data folder, to mount persistent volume to store locale files(at least en.json and locales.json) for offline installations.


NGPixel:
I suggest using the CONFIG_FILE env variable instead to define the path. This way the config file can be located anywhere, no symlink needed.

PR #1269, opened 2019-11-27
FIX Postgres Engine to query using right locale

vibornoff:
This fixes search queries with russian locale for me.

### DIFF
--- a/server/modules/search/postgres/engine.js
+++ b/server/modules/search/postgres/engine.js
@@ -62,7 +62,7 @@ module.exports = {
       let suggestions = []
       const results = await WIKI.models.knex.raw(`
         SELECT id, path, locale, title, description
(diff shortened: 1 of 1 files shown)


NGPixel:
I modified your code to use parameter binding instead of SQL injection

vibornoff:
You should probably do that all over the `server/modules/search/postgres/engine.js`:
- [L91](https://github.com/Requarks/wiki/blob/master/server/modules/search/postgres/engine.js#L91)
- [L105-L107](https://github.com/Requarks/wiki/blob/master/server/modules/search/postgres/engine.js#L105-L107)
- [L155](https://github.com/Requarks/wiki/blob/master/server/modules/search/postgres/engine.js#L155)

PR #1343, opened 2019-12-30
Allow ca cert for mysql/maria db connection

UnbrandedTech:


NGPixel:
Implemented in f1725159f76b95fd296f684db9e25e76f59dcb9d

Changed your code to allow any SSL option supported by nodejs and to work on postgresql connections as well.

PR #1395, opened 2020-01-19
Added Env-Variable for logLevel

InterFelix:
Added Env-Variable LOG_LEVEL to set logLevel in config.yaml on Docker-based installations.

Disclaimer: I don't know if this edit could have unintended consequences for other installation options. Could only test it on Docker.

### DIFF
--- a/dev/build/config.yml
+++ b/dev/build/config.yml
@@ -15,4 +15,4 @@ ssl:
   provider: letsencrypt
   domain: $(LETSENCRYPT_DOMAIN)
   subscriberEmail: $(LETSENCRYPT_EMAIL)
-logLevel: info
+logLevel: $(LOG_LEVEL)


InterFelix:
Okay. This 'fix' doesn't seem to work for me. I can't get the server to log anything else than info. I need verbose logging to troubleshoot my LDAP login stuff, but no matter what I try, I can't get any information on why the authentication fails out of the logs.
Could you look into making logLevel available as environmental variable for docker?

NGPixel:
Have you read the docs first?
https://docs.requarks.io/en/auth/ldap#troubleshooting

InterFelix:
Thanks for your suggestion. I enabled the flag, but the output doesn't give me any more information than is already displayed in the interface. I believe it's the exact same words. It just says "LDAP LOGIN ERROR (c1): Invalid email / username or password."
Not helpful at all. However, I could not test the flag with different logLevels.

NGPixel:
You should get an error "LDAP LOGIN ERROR (c2)” in the logs. (c1) is the fallback error.

InterFelix:
I only get a (c1) error. No (c2) for me. But I won't put any more effort into this LDAP-thing, our SSO will soon be switching to OpenID Connect anyways. And maybe wiki.js also supports it by then, because we also have some development to do until we can switch.

Issue #1294, opened 2019-12-09
Error while importing content from Wiki.js 1.x.x or editing pages with long content


SConstantinou:
**Describe the bug**
Wiki.js 2.0.12 with DB mysql
table pageHistory column content is of type text
this causes problems when importing from Wiki.js 1.x or when using the "Import Everything" action under storage. For most entries in the table we get warnings like: "Data too long for column 'content' at row 1". The behaviour is the same when editing a pages and trying to add long content.

**To Reproduce**
Import pages with long content (over 64KB) especially entered all at once.

**Expected behavior**
Page should be updated normally without error. LONGTEXT should be used in mysql instead of text.

**Screenshots**

**Host Info (please complete the following information):**
 - OS: Ubuntu 18.04
 - Wiki.js version: 2.0.12]
 - Database engine: MariaDB 10.4.10]

**Additional context**
nsert into `pageHistory` (`action`, `authorId`, `content`, `contentType`, `createdAt`, `description`, `editorKey`, `hash`, `isPrivate`, `isPublished`, `localeCode`, `pageId`, `path`, `publishEndDate`, `publishStartDate`, `title`) values ('updated', 1,

.......

Data too long for column 'content' at row 1"

SConstantinou:
After you get the error, then you are not able to do anything with the page as you cannot even delete content or the whole page.

NGPixel:
The pages table is actually using LONGTEXT. however the history table is still set with normal TEXT type, which is a bug.

PR #1400, opened 2020-01-21
fix: alter pageHistory content data type to LONGTEXT (#1294)

fireundubh:
Per @NGPixel in #1294, the column data type for `pageHistory.content` should be initialized as `LONGTEXT`. This fix affects only new installs and new migrations. For existing installs, either wiki admins will be expected to fix the column type in their databases, or a "validate and fix schema" option will need to be added to the Utilities menu.

### DIFF
--- a/server/db/beta/migrations/2.0.0-rc.29.js
+++ b/server/db/beta/migrations/2.0.0-rc.29.js
@@ -15,6 +15,17 @@ exports.up = knex => {
           break
       }
     })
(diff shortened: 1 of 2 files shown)


NGPixel:
This change should be done in a new migration, not by modifying an existing one.

NGPixel:
Fixed by 4dc7e0537306f78a7b3a6c27c64de02615bf34c2

PR #1283, opened 2019-12-02
Fix list item formatting for lists with paragraphs

nilium:
This removes the fancy tick-mark next to unordered list items, which is
a little depressing, but fixes formatting for them when they contain
paragraphs (`<p>` elements). The tick-mark has been replaced with
a square mark. Both ordered and unordered lists have grown 1rem of left
padding to compensate for their position.

Considered using disclosure-closed as a similar triangle replacement,
but it's not regarded as something that should be used in production
(per MDN). Using a simple string is unsupported because it's only in
a CSS working draft, so a square is the least controversial option,
I think.

Closes #1282.

PR #1406, opened 2020-01-22
Fix list formatting multiline with fancy bullets

takashi1kun:
In kinda response to: Requarks#1283 because i kinda like the fancy bullets so i propose a alternative solution that solves the problem with the list having issues with p elements and with multi lines but keeping the fancy css bullets.

Tested on Chrome, Firefox, Opera and Edge.
image of it working on chrome:
chrome_2020-01-22_14-55-47
and image of it working on firefox:
firefox_2020-01-22_15-11-36

I did not include screenshots of the other because i believe that showing this may be enough, but i can provide it if needed

I have 2 alternative css for the p like 

```scss
  > p {
        display:inline-table;
      }
```
or
```scss
  > p {
        display:inline-flex;
      }
```
i just choose to use 
```scss
  > p {
        display:inline-block;
        vertical-align:top;
        padding-top:0!important;
      }

```
because i believe that it has more cross browser compatibility, but science for the other fixes i already had to use flex it may not matter.

### DIFF
--- a/client/themes/default/scss/app.scss
+++ b/client/themes/default/scss/app.scss
@@ -487,7 +487,7 @@ ul {
       > p {
         display:inline-block;
         vertical-align:top;
-        padding-top:0!important;
+        padding-top:0;
       }
   }
   }
--- a/client/themes/default/scss/app.scss
+++ b/client/themes/default/scss/app.scss
@@ -475,11 +475,15 @@
 ul {
     list-style: none;
     width: 100%;
-    > li::before {		
-      content: '\25b8';		
-      color: mc('grey', '600');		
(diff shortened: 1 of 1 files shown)


takashi1kun:
as a plus this solves the problem caused with link-lists having bullets mentioned here 

> https://github.com/Requarks/wiki/pull/1283#issuecomment-566330473

image

takashi1kun:
I have to add that only with the bit of code that affects the p element is enough to solve the bug with p elements, the code affecting the li and ul solve a bug when you add a multiline to a list pressing enter and no p element is generated surronding the code.

PR #1429, opened 2020-02-01
Provide a simple working helm chart

Nox-404:
Hello,
here is a draft for a working helm chart.

There are a few limitation for now but if needed I can update it further.

### DIFF
--- a/dev/helm/config/config.base.yml
+++ b/dev/helm/config/config.base.yml
@@ -1,0 +1,136 @@
+#######################################################################
+# Wiki.js - CONFIGURATION                                             #
+#######################################################################
+# Full documentation + examples:
+# https://docs.requarks.io/install
+
+# ---------------------------------------------------------------------
+# Port the server should listen to
+# ---------------------------------------------------------------------
+
+port: 3000
+
+trustProxy: $(TRUST_PROXY)
+
+# ---------------------------------------------------------------------
+# Database
+# ---------------------------------------------------------------------
+# Supported Database Engines:
+# - postgres = PostgreSQL 9.5 or later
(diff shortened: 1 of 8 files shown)


NGPixel:
I changed the postgres requirement version to 11 (unless there was a specific reason for using 8.2.1?)

Let me know if you want to make any other changes before I merge it.

Nox-404:
8.2.1 is the latest `chart` version not the actual `postgresql version`. Version 11 of the chart gives a not found on dependencies

Nox-404:
The `8.2.1` postgresql chart uses `bitnami/postgresql:11.6.0-debian-10-r5`

Nox-404:
Also there are still some pitfalls:
* [ ] Stuck with postgresql db backend
* [ ] No persistent volume for `local` storage strategy
* [ ] SSL can only be handled at the `ingress-controller` level (might be tweaked using annotations and stuff, but not that easy)
* [ ] No documentation yet
* [ ] No package, test & release provided (I have no exp with that)

Let me know if you what you want for the helm and I'll do it when I have some time :)

NGPixel:
Ah I see, my bad. I can definitely help with the packaging / release part.

Having only postgres is not really an issue IMO. I'd rather have everyone use postgres if they're gonna deploy it on a Kubernetes cluster. As for SSL, I'd assume most people would handle it at the ingress level anyway.

The persistent volume could be a nice to have, but again, not a requirement (a remote git is usually a better idea).

Feel free to submit a PR for this file for the documentation:
https://github.com/Requarks/wiki-docs/blob/master/install/kubernetes.md

Thanks for your help!

PR #1430, opened 2020-02-01
Code Block in Info Block Matches Other Code Blocks

### LINKED ISSUE #1411: Code Block inside info block blurry
**Describe the bug**
When there is a code block inside of an info block, the text in the code block becomes blurred unless highlighted.

**To Reproduce**
1. In the Markdown text editor, create an info block
2. Create a code block inside the info block
3. Save and view the page
4. See that the code block is blurry

**Expected behavior**
The code block text should not be blurry when inside of an info block.

**Screenshots**

Shows the code block inside an info block being blurry.
unknown

Shows the code block inside an info block with clear text when highlighted, and compared to a code block outside the info block.
unknown-2

 - OS: DigitalOcean Marketplace release, Ubuntu 18.04?, docker.
 - Wiki.js version: 2.0.12
 - Database engine: postgres

**Additional context**
Using Cloudflare as a CDN (however the js & performance features have been disabled since it causes issues with Subresource Integrity, even when turned off in the admin panel).

lilnayners:
Fixes #1411

### DIFF
--- a/client/themes/default/scss/app.scss
+++ b/client/themes/default/scss/app.scss
@@ -245,11 +245,6 @@
       border-color: mc('blue', '300');
       color: mc('blue', '900');
 
-      code {
-        background-color: mc('blue', '50');
-        color: mc('blue', '800');
-      }
-
       @at-root .theme--dark & {
         background-color: mc('blue', '900');
         color: mc('blue', '50');


NGPixel:
The code block you just removed is meant for inline code styling *(see this example where the inline code text `latest` is blue: https://docs.requarks.io/en/install/docker#using-the-docker-image)*

Code blocks will actually need extra css to override that behavior using additional selector `.prismjs > code`.

lilnayners:
Ah ok, I see, that makes sense. So the inline code should be styled differently than the block code. So should the code blocks within info boxes look the same as other code blocks with the black background? I just want to make sure before making further changes.

NGPixel:
Correct.

PR #1432, opened 2020-02-03
Adding an override for the ssh port

wallrick:
Problem:
- Trying to use git storage feature with a private git server with ssh running on a non-default port.  Git over ssh typically happens on port 22 however non-standard setups can use a custom port number.  Wiki.js doesn't support adding custom SSH port while setting up the git storage.

Solution:
- Added an option on the git storage configuration page to allow specifying ssh port.  It is optional and empty by default with port 22 used by default if nothing specified.

Testing:
- Manually validated that the setting is properly picked up.
- Tested with custom and default ports and everything worked as expected.

Possible future improvement:
- Adding an advanced ssh section on the git storage page and rolling this configuration into that would be a possible way of simplifying the UI.

### DIFF
--- a/server/modules/storage/git/definition.yml
+++ b/server/modules/storage/git/definition.yml
@@ -94,11 +94,11 @@ props:
     hint: Optional - Absolute path to the Git binary, when not available in PATH. Leave empty to use the default PATH location (recommended).
     order: 50
   sshPort:
-    type: Integer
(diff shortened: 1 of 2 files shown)


craviee:
This was removed somehow? I'm using version 2.5.170 which should have it but it doesn't.

PR #1447, opened 2020-02-06
Fix/page rules role check

### LINKED ISSUE #1425: Cannot set page-specific write permissions
I want some pages to be read-only and other pages to be writable. I set everything up as seen in the screenshot:

permissions

- the first rule is to start by denying everything
- then I allow specific access to the homepage the users are directed to when logging in (read-only)
- I allow full access to the users readwrite-area
- I allow read-only to a special immutable area.

For reading this works as expected (e.g. other existing pages cannot be opened). For writing only the global content permissions seem to have any effect. If I disable global write, the user cannot write on any pages, no matter what the page-specific permissions are. If I enable global write, the user can write to all pages they have read permissions on, no matter what other page-specific permissions are set. Any help would be greatly appreciated. Not sure if this a bug or a lack of understanding on my part.

bobonthenet:
This should resolve #1425

### DIFF
--- a/server/core/auth.js
+++ b/server/core/auth.js
@@ -152,7 +152,7 @@ module.exports = {
    */
   checkAccess(user, permissions = [], page = false) {
     const userPermissions = user.permissions ? user.permissions : user.getGlobalPermissions()
-
+    const checkPermissions = permissions
     // System Admin
     if (_.includes(userPermissions, 'manage:system')) {
       return true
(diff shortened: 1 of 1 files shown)


server/core/auth.js
```diff
@@ -152,7 +152,7 @@ module.exports = {
    */
   checkAccess(user, permissions = [], page = false) {
     const userPermissions = user.permissions ? user.permissions : user.getGlobalPermissions()
-
+    const checkPermissions = permissions
```
nicolasg-playster: Use `permissions` directly instead of creating a new copy. `_.intersection` does not mutate the original, so it's perfectly safe to use it directly.

server/core/auth.js
```diff
@@ -173,26 +173,28 @@ module.exports = {
       user.groups.forEach(grp => {
         const grpId = _.isObject(grp) ? _.get(grp, 'id', 0) : grp
         _.get(WIKI.auth.groups, `${grpId}.pageRules`, []).forEach(rule => {
+          let hasRole = _.intersection(rule.roles, checkPermissions).length > 0
+
           switch (rule.match) {
             case 'START':
-              if (_.startsWith(`/${page.path}`, `/${rule.path}`)) {
+              if (_.startsWith(`/${page.path}`, `/${rule.path}`) && hasRole) {
```
nicolasg-playster: Move `hasRole` at the beginning of the expression, so it's faster to exit the condition if false. ( repeat for other rules below )

server/core/auth.js
bobonthenet: For some reason I thought there was going to be a scoping issue. I just retried it though and I was mistaken.

server/core/auth.js
bobonthenet: I got rid of hasRole all together and wrapped the entire switch in an if statement that checks for the role in the permission with _.intersection.

dakira:
Thanks a lot for this PR!

PR #1524, opened 2020-02-24
Admin -> Pages -> Edit Path -> Save/Update

CubE135:
**Reasoning for Pull Request:**
As i was confused by the edit icons for the path and locale when editing a page, i decided to add some function to it.

**Changes:**
For now, i added a "coming soon" notification to the locale edit icon and focused on the path editing only. But i prepared a bit for adding locale editing in the near future.
When clicking the path edit icon, the user will be presented with a input field. The user can then edit the path to his liking.
After editing the path, the user will need to save the changes. I added a "Update Page" Button to the top right of the page.
Pressing that button will save the changed path to the database.
For editing other values on that page, it is now very easy to add more editable values in the future.

**Code clarification:**
For some reason, when trying to update the page through a mutation, i was told that the page content can not be empty. That doesnt really make sense when editing only the path, so i made it possible to edit the page without providing page-content.

I also ran into an error, when editing the page, because of the tags.
Sorrounding the `associateTags-call` with a simple "if" solved the issue.

**Locales:**
As i had to add some new texts, these are the language keys i added:
```
common:
  page.updatePage: 'Update Page'
  page.pageUpdateSuccess: 'Page updated successfully.'
```

Best Regards,
CubE :)

### DIFF
--- a/client/components/admin/admin-pages-edit.vue
+++ b/client/components/admin/admin-pages-edit.vue
@@ -189,7 +189,6 @@ import _ from 'lodash'
 import { StatusIndicator } from 'vue-status-indicator'
 import gql from 'graphql-tag'
 
-import localesQuery from 'gql/admin/locale/locale-query-list.gql'
 import pageQuery from 'gql/admin/pages/pages-query-single.gql'
 import deletePageMutation from 'gql/common/common-pages-mutation-delete.gql'
 
@@ -202,17 +201,11 @@ export default {
       deletePageDialog: false,
       page: {},
       loading: false,
-      locales: [],
       selectedLocale: 'en',
       editPop: {
         path: false,
         timezone: false
       }
-    }
-  },
-  computed: {
-    installedLocales() {
-      return _.filter(this.locales, ['isInstalled', true])
     }
   },
   methods: {
@@ -272,7 +265,7 @@ export default {
      * Update the current page
      */
     async updatePage() {
-      this.$store.commit(`loadingStart`, 'admin-users-update')
+      this.$store.commit(`loadingStart`, 'page-update')
       try {
         let resp = await this.$apollo.mutate({
           mutation: gql`
@@ -317,7 +310,7 @@ export default {
         })
         throw err
       }
-      this.$store.commit(`loadingStop`, 'admin-users-update')
+      this.$store.commit(`loadingStop`, 'page-update')
     }
   },
   apollo: {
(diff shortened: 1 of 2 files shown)


client/components/admin/admin-pages-edit.vue
```diff
@@ -165,7 +187,9 @@
 <script>
 import _ from 'lodash'
 import { StatusIndicator } from 'vue-status-indicator'
+import gql from 'graphql-tag'
 
+import localesQuery from 'gql/admin/locale/locale-query-list.gql'
```
NGPixel: You actually don't need to remotely fetch the installed locales, you have them already in the `siteLangs` variable, which is available everywhere.

Example value:
```
[
  {
    "code":"en",
    "name":"English"
  },
  {
    "code":"fr",
    "name":"français"
  }
]
```

client/components/admin/admin-pages-edit.vue
```diff
@@ -212,6 +247,77 @@ export default {
...
+        _.delay(() => {
+          this.$refs[ipt].focus()
+        }, 200)
+      })
+    },
+    /**
+     * Update the current page
+     */
+    async updatePage() {
+      this.$store.commit(`loadingStart`, 'admin-users-update')
```
NGPixel: Make sure all the loading triggers have a proper ID.

NGPixel:
Maybe calling the pages > move mutation would make more sense, rather than update:
https://github.com/Requarks/wiki/blob/master/server/graph/schemas/page.graphql#L95-L99

client/components/admin/admin-pages-edit.vue
CubE135: Oh ok, i will fix that :)

client/components/admin/admin-pages-edit.vue
CubE135: True, will be fixed :)

CubE135:
> Maybe calling the pages > move mutation would make more sense, rather than update:
> https://github.com/Requarks/wiki/blob/master/server/graph/schemas/page.graphql#L95-L99
I thoght, the feature may be extended at some point in time, where other values on that page would be editable too. Then, we would need the update mutation anyways, right?

NGPixel:
Good point.

CubE135:
Fixed the things you mentioned :)

CubE135:
Hey :) Is there anything else i need to fix?

PR #1528, opened 2020-02-25
Render "index" pages (index.md, README.md)

supportpartnership:
Adopting "mkdocs"-based sites is sub-optimal, as mkdocs supports "index" pages that get automatically served from subdirectories: https://www.mkdocs.org/user-guide/writing-your-docs/#index-pages

This patch attempts to load a page directly from its path and returns it if exists, but if it doesn't then it attempts to load the page from some pre-defined locations. This makes adopting mkdocs-based sites a breeze.

This could probably do with some options to better define the default pages, but in my tests it works quite well.

### DIFF
--- a/server/models/pages.js
+++ b/server/models/pages.js
@@ -665,12 +665,12 @@ module.exports = class Page extends Model {
   }
 
   /**
-   * Fetch an Existing Page from the Database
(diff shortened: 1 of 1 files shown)


NGPixel:
This should be done at indexing time from the storage module, not on every page load.

PR #1548, opened 2020-03-01
Option to restrict Discord auth to a specific server

lightspeedlucas:
Very small feature to optionally restrict discord authentication to members of a specific server. We use this in our wiki, maybe it could be useful to others.


Jhobean:
Nice feature!

Issue #1628, opened 2020-03-26
"Visualize Pages" SVG does not fit to width, is not horizontally scrollable


eeshugerman:
**Describe the bug**
I'm am unable to view the entirety of the "Visualize Pages" SVG, because it does not fit to the width of my browser window and the page is not horizontally scrollable.

**To Reproduce**
1. Create a largish hierarchy of pages
2. Click Administration -> Pages -> VISUALIZE -> Hierarchical Tree
3. Observe

**Expected behavior**
I am able to view the full SVG.

**Screenshots**
![image](https://user-images.githubusercontent.com/25518211/77696424-f8d6c100-6f72-11ea-82c9-500f5e5e86cc.png)

----

If I make my browser window narrower, the problem is even more apparent:

![image](https://user-images.githubusercontent.com/25518211/77696613-46532e00-6f73-11ea-8edc-331191602f57.png)

**Host Info (please complete the following information):**
 - OS: Docker
 - Wiki.js version: 2.1.113
 - Database engine: PostgreSQL 12

PR #1634, opened 2020-03-30
fix: don't clip "visualize pages" svg

eeshugerman:
closes https://github.com/Requarks/wiki/issues/1628

i am far from a front-end dev, so please let me know if there's a better way to do this

### DIFF
--- a/client/components/admin/admin-pages-visualize.vue
+++ b/client/components/admin/admin-pages-visualize.vue
@@ -371,9 +371,10 @@ export default {
 <style lang='scss'>
 .admin-pages-visualize-svg {
   text-align: center;
-
+  overflow-x: auto;
   > svg {
     height: 100vh;
+    width: 80vw;
   }
 }
 </style>


eeshugerman:
image

eeshugerman:
So the `width: 80vw` bit is pretty much to account for the sidebar. I didn't love this in the first place ("no magic numbers"), but now that I see https://github.com/Requarks/wiki/commit/a7de0f020fdf716544ad9dbf513ac95defb714a1, I would definitely like to find a better way to do this.

IIRC I tried `width: 100%` which, from what I can tell, is relative to the parent container (good), but then scrolling was a never activated -- even really "big" SVGs are just scaled down to the size of the page.

Maybe it would be best to just not set the width. But then scrolling is often activated when it isn't really necessary (ie the SVG could be viewed comfortably if scaled down a bit to fit the screen). Still, it's simple and would be an improvement.

Maybe adjusting `viewBox` would be part of the proper solution? I haven't experimented with that at all.

Also it seems there's something else at play -- even with scrolling activated, sometimes part of the SVG is still clipped. Example:

image

eeshugerman:
I think the best way to do this would be to fit the SVG to width/height of the screen, so scrolling is not necessary but make it zoomable and pannable. I'm taking a stab at this with `d3-zoom`, but am having trouble getting it just right. Hopefully someone more familiar with front end development can take this on.

eeshugerman:
UPDATE: Got all the kinks worked out with `d3-zoom`! Included here: https://github.com/Requarks/wiki/pull/1914

Issue #1638, opened 2020-03-30
Can't use Ctrl-Save to close the editor on a Mac


sbonaime:
**Describe the bug**

Ctrl-Save to close the editor on a Mac does only save. I tried with Firefox and Chrome and also Command-Save and option-Save

Some other users are reporting this bug
https://requarks.canny.io/wiki/p/cant-use-ctrl-save-to-close-the-editor-on-a-mac

**Host Info (please complete the following information):**
 - Wiki.js version:2.2.51

sbonaime:
CMD + Click the create button (in a new page) does save and does close !
Option + Click : Save only
CMD + Click : Save only

@NGPixel in 
https://github.com/Requarks/wiki/blob/master/client/components/editor.vue

Maybe that  @click='save' is trigger before @click.ctrl.exact='saveAndClose'
what about changing it to 

 @click.exact='save'

Smankusors:
well can you help us by trying to build wiki with your patch to see if it really works on Mac?

(sorry, but you used "maybe" makes me doubt if this works tho)

sbonaime:
@Smankusors I would love to do it but right now, I am not able to build wiki because I am stuck at @home   :-(

sbonaime:
I manage to install the dev. 
From my tests on my Mac, Cmd is the main modifier used by wiki.js on Mac : 
- Cmd + S : Save the page
- Cmd + B : Bold
- ....

BUT

Cmd + "save button" does not work because the code is '@click.ctrl='saveAndClose'. By the way, Ctrl+ "save button" does not work either something I don't really understand. 

I manage to make it working by using meta (or alt or shift) modifier like that in 
https://github.com/Requarks/wiki/blob/master/client/components/editor.vue

          @click='save'
          @click.meta='saveAndClose'

But I think it is breaking the Windows and maybe Linux functionality. If we use meta, the windows user will need to use "Windows key"+click instead of "ctrl"+click.

So maybe the modifier should be filtered. I found this

NGPixel:
A dedicated save and close button was added in the upcoming v3.

PR #1641, opened 2020-03-30
Maybe solve #1638

sbonaime:
Maybe that @click='save' is trigger before @click.ctrl.exact='saveAndClose' on Macos


sbonaime:
after testing, I did not solve the problem :-(

PR #1663, opened 2020-04-02
Acitvated scrolling for page tree visualisation

sray:
In chrome, the tree visualisation is scrollable on the y axis, but not on the x-axis. This PR makes the tree visualisation scrollable on the x-axis as well. This is a quick fix, there might be a better position and a more suitable place for the css than proposed with this pull request. If this is the case, let me know or simply edit it.


NGPixel:
Duplicate of #1634

PR #1751, opened 2020-04-22
Update helm chart configs and readme

fuziontech:
Update helm chart configs and readme. Tested on local kubernetes cluster with nfs pv class.


dev/helm/templates/deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+              containerPort: 3000
+              protocol: TCP
+          livenessProbe:
+            httpGet:
+              path: /
+              port: http
+          readinessProbe:
+            httpGet:
+              path: /
+              port: http
```
NGPixel, line 63: The `/healthz` endpoint would be better suited.

```suggestion
          livenessProbe:
            httpGet:
              path: /healthz
              port: http
          readinessProbe:
            httpGet:
              path: /healthz
              port: http
```

don-sh:
Shouldn't there be an option to set the DB_TYPE if postgres is disabled?

NGPixel:
> Shouldn't there be an option to set the DB_TYPE if postgres is disabled?

No, as PostgreSQL is mandatory for high-availability.

don-sh:
> > Shouldn't there be an option to set the DB_TYPE if postgres is disabled?
> 
> No, as PostgreSQL is mandatory for high-availability.

Then you need to remove the {{ if .Values.postgresql.enabled }} below - otherwise you run into an "invalid DB_TYPE".
```
            {{ if .Values.postgresql.enabled }}
            - name: DB_TYPE
              value: postgres
            {{ end }}
```

igorfernandes:
Any update about this PR?

fuziontech:
need to update with feedback. I'll do that tonight

fuziontech:
@don-sh - You are right, updated `deployment.yaml` to only support postgres and not remove that config if `postgres.enabled` is set false

PR #1765, opened 2020-04-23
add env variables for custom certificate usage

UnlimitedCookies:
This would make it possible to use a custom certificate and therefore we gain more flexibility.
Before using a custom certificate, first it must be mounted with a volume. (e.g. with a volume entry  
like ` - /etc/ssl/wikijs:/path/to/folder/on/host:ro` in the Docker Compose file).
Unfortunately this is **not fully backwards compatible**, because the `SSL_PROVIDER` env variable has been added, so that all users currently using LE need to set this to `letsencrypt`. Maybe letsencrypt could be used as default so that old environment variables still work. Let me know if you have any better suggestions.
I would update the wiki accordingly if this PR is being considered.

**The following docker environment variables have been added:**
`SSL_PROVIDER` - This is either letsencrypt or custom
`SSL_FORMAT` - This is either pem or pfx
`SSL_PEM_KEY_FILEPATH` - File path to the key file in PEM on the container
`SSL_PEM_CERT_FILEPATH` - File path to the cert file in PEM on the container
`SSL_PFX_FILEPATH` - File path to the combined cert/key file in PFX on the container
`SSL_PASSPHRASE` - Passphrase for the Key files
`SSL_DHPARAM` - Diffie Hellman parameters

Please let me know what you think.

### DIFF
--- a/dev/build/config.yml
+++ b/dev/build/config.yml
@@ -12,7 +12,13 @@ db:
 ssl:
   enabled: $(SSL_ACTIVE)
   port: 3443
-  provider: letsencrypt
+  provider: $(SSL_PROVIDER)
+  format: $(SSL_FORMAT)
+  key: $(SSL_PEM_KEY_FILEPATH)
+  cert: $(SSL_PEM_CERT_FILEPATH)
+  pfx: $(SSL_PFX_FILEPATH)
+  passphrase: $(SSL_PASSPHRASE)
+  dhparam: $(SSL_DHPARAM)
   domain: $(LETSENCRYPT_DOMAIN)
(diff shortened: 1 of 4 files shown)


NGPixel:
`letsencrypt` should indeed be the default to stay fully backward compatible.

Same thing with `SSL_FORMAT` which should default to `pem`

UnlimitedCookies:
I've added a file change for defaulting to letsencrypt. Although this should already work, the systeminfo functions still need to default to letsencrypt.

Why should the `SSL_FORMAT` default to `pem` though, since this always had to be specified before?

PR #1829, opened 2020-05-04
make TOC scrollable if higher than the viewport

Pwuts:
Erratum on #1825 

**Problem:**
If the sticky table of contents is higher than the viewport, the overflowing part can only be accessed by scrolling to the bottom of the page.

**Solution:**
Make the table of contents itself scrollable and limit its height to that of the viewport.


Pwuts:
image
Screenshot shows a version with scrollbar, but imo the scrollbar was too prominent that way so I left it out in the PR.

Pwuts:
resolved conflict

Pwuts:
@NGPixel I would advise processing this PR before the next release, as it should have been included in #1825 which has already been merged

PR #1865, opened 2020-05-11
Page-component feature with components in /Components.

codevin:
This is a feature to allow vue components to be specified in wiki pages. Response to feature request "Dynamic tables". Opens up many possibilities, for e.g. create tables of data from remote websites. I can give examples of wiki pages, but not sure where to include them since we don't have docs.
Also, not sure how to make this as optional feature that can be enabled in config.

I have included axios and bootstrap-vue as dependencies; it ensures we don't have to struggle with imports. 

This needs to be reviewed and discussed.


codevin:
Forgot to add this in documentation: We need safe-html to be off for it to work. 
In the longer run, we should have some easier mechanism to allow tags with some format, such as starting with 'pgc-'  to be recognized and allowed in html-security plugin.

codevin:
Another issue I noticed in recent production versions is that 'v-pre' tag is used to wrap the page content (in ./server/views/page.pug). That would mean the page components are not rendered. 
Need to accommodate this in some way.

NGPixel:
This is no longer the case in 2.4.

codevin:
I presume 'master' is 2.4? There is no separate branch or tag for 2.4 yet, right?

codevin:
Currently user without 'manage' or 'delete' access to graphql query on page is unable to read from the page. I think it is a mistake. Anyone who can read the page should be able to do 'single' page call from graphquery. I have opened issue #1895 for this.

Maxzor:
My 2cents as a big noob in webdev, is that wikimedia made a fundamentally wrong design decision with wikitext, which is so garbage that virtually all the content ever produced in wikipedia is forever irrecoverable.

I am exaggerating, but I don't think having vue components defined in userspace html is a great idea.
It would be in same the trend of making the page contents not reusable elsewhere without significant effort.

If the user is tech-savvy enough and wants a custom component, make her/him add it as a wikijs _extension_ and call it from the page content

> #1865 with 'pgc-'

 / some syntactic sugar

This will also encourage contributions.

PR #1899, opened 2020-05-17
Added Rocket.Chat authentication

jschirrmacher:
I liked to authenticate at Wiki.js with my Rocket.Chat server via OAuth2, so I created this authentication module and would be happy if you would use it in the standard product as well.


NGPixel:
I would avoid creating additional modules which are simply a standard OAuth2 module. The generic OAuth2 module is coming in the 2.5 update.

Unless there's something really specific in a Rocket.Chat implementation that is not OAuth2-compliant, using the generic module would be preferred.

jschirrmacher:
Great to hear, that the generic auth module gets into 2.5! I would very much appreciate that.
Will a mapping of user attributes be part of that? If not, it would still be necessary to have a separate module for that, I think.
When will 2.5 be available, wdyt?

PR #1905, opened 2020-05-19
Development docker-compose.yml improvements

eeshugerman:
- Adds `container_name` fields for cli convenience
- Removes the `wikinet` network -- Compose creates a network automatically
- Bind-mounts selectively
  - This fixes issues with the Git storage module in dev/docker mode. The issue is `/wiki/.git/` is bind mounted to the repo on the host, which interferes with the app's Git operations.

Let me know if you'd like me to rename commits (fix? feat?) or break it up into multiple PRs or anything like that.

### DIFF
--- a/dev/containers/docker-compose.yml
+++ b/dev/containers/docker-compose.yml
@@ -4,6 +4,7 @@
 version: "3"
 services:
   db:
+    container_name: wiki-db
     image: postgres:9-alpine
     environment:
       POSTGRES_DB: wiki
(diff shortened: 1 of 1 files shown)


dev/containers/docker-compose.yml
```diff
@@ -43,21 +40,22 @@ services:
...
       - "3000:3000"
     volumes:
-      - ../..:/wiki
-      - /wiki/node_modules
+      - ../../client:/wiki/client
+      - ../../server:/wiki/server
+      - ../../dev:/wiki/dev
+      - ../../package.json:/wiki/package.json
+      - ../../yarn.lock:/wiki/yarn.lock
+      - ../../.babelrc:/wiki/.babelrc
```
NGPixel, line 57: You're missing a few files here. All files at the root of the project should also be mounted.

eeshugerman: Ah true, I thought the others were unnecessary but I see now that they won't appear in the VS Code project unless listed here. Bummer, was hoping to avoid that. I'll try
```
- ../..:/wiki
- /wiki/node_modules
- /wiki/.git

eeshugerman: Cool, the above works. Cleaner than my initial solution anyway!

Issue #1604, opened 2020-03-20
S3 Export all don't work


rlarrode:
**Describe the bug**
In s3 Storage Admin section, when i click on export all, i get success popup but nothing happen on s3 side.
On server side i get following error
`[MASTER] warn: Invalid Handler for Storage Target`

When i manually update a page, it is successfully saved in my S3

**To Reproduce**
1. Provide S3 Storage in Administration Panel
2. Click on "Export All"
3. Check your S3 bucket (nothing change)
4. See error in server logs

**Expected behavior**
All DB content should be transfered to S3 bucket

**Host Info:**
 - OS: Debian 9.0 on AWS EC2
 - Wiki.js version: 2.1.113
 - Database engine: postgres 9.5

Thanks

Blizarre:
I don't know if this might help, but I get the same problem.
On top of that, when I go the admin storage page, I get a popup `An unexpected error occured.`. The response to the call to the GraphQL endpoint made by the browser contains this error:
```
        "extensions": {
          "code": "INTERNAL_SERVER_ERROR",
          "exception": {
            "stacktrace": [
              "Error: Cannot return null for non-nullable field StorageStatus.message.",
              "    at completeValue (/wiki/node_modules/graphql/execution/execute.js:560:13)",
              "    at completeValueCatchingError (/wiki/node_modules/graphql/execution/execute.js:495:19)",
              "    at resolveField (/wiki/node_modules/graphql/execution/execute.js:435:10)",
              "    at executeFields (/wiki/node_modules/graphql/execution/execute.js:275:18)",
              "    at collectAndExecuteSubfields (/wiki/node_modules/graphql/execution/execute.js:713:10)",
              "    at completeObjectValue (/wiki/node_modules/graphql/execution/execute.js:703:10)",
              "    at completeValue (/wiki/node_modules/graphql/execution/execute.js:591:12)",
              "    at completeValueCatchingError (/wiki/node_modules/graphql/execution/execute.js:495:19)",
              "    at /wiki/node_modules/graphql/execution/execute.js:618:25",
              "    at Array.forEach (<anonymous>)",
              "    at forEach (/wiki/node_modules/iterall/index.js:83:25)",
              "    at completeListValue (/wiki/node_modules/graphql/execution/execute.js:614:24)",
              "    at completeValue (/wiki/node_modules/graphql/execution/execute.js:573:12)",
              "    at /wiki/node_modules/graphql/execution/execute.js:492:16",
[...]

PR #1922, opened 2020-05-22
fix: S3 Export all don't work #1604

regevbr:
fix: #1604

exportAll action on s3 doesn't work.
The reason being that s3 handler is a class instance and not a plain object, thus when checking if the handler has the exportAll method, fails.
The fix is to use lodash `hasIn` which checks if a propery exists directly or inherited on an object.

I confirmed the fix locally ofcourse


regevbr:
@NGPixel can you please CR this?

regevbr:
Thanks @NGPixel  :-)

PR #1924, opened 2020-05-23
Initial support for captioned figures with optional floating

despens:
In markdown, users can use the class `caption` below an image to create a figure with an integrated caption, in the same width as the image. (The image might need to be manually sized if it is too large.) Optionally, the figure can be floated left or right using additional classes `left` or `right`.

Example with caption on the same line:

```markdown
![A nice photo](/image.jpg) Photo taken by my friend Bob.{.caption .left}
```

Example with caption in the following paragraph and manual image size:

```markdown
![A nice photo](/image.jpg =300x)

Photo taken by my friend Alice.{.caption .right}
```

Example screen shot:

image


NGPixel:
A few things to improve:

- Your CSS should be part of ./client/themes/default/scss/app.scss as it's content styling.
- Avoid too specific CSS selectors (`.illustration` instead of `div.illustration`)
- Use SCSS encapsultation instead of redefining the whole path for each CSS block.
- Don't use `.caption` CSS class name as it's already used by Vuetify.
- Use `.align-left` and `.align-right` for alignment class names, to stay consistent with the already defined alignment classes for images.

server/modules/rendering/html-figcaption/renderer.js
```diff
@@ -0,0 +1,66 @@
...
+        figurable_elm = elm.children(figurable_selector)
+        caption_elm = $('<figcaption>')
+          .append(elm.contents())
+        replace_elm = elm
+        transform = true
+      }
+      // figurable to be captioned is in the paragraph above
+      else {
+        let prev = elm.prev('p')
+        if (prev.text().trim().length == 0) {
```
NGPixel, line 38: Use === instead

server/modules/rendering/html-figcaption/renderer.js
```diff
@@ -0,0 +1,66 @@
...
+      // figurable to be captioned is in the same paragraph
+      if (elm.contents(':first-child').is(figurable_selector)) {
+        figurable_elm = elm.children(figurable_selector)
+        caption_elm = $('<figcaption>')
+          .append(elm.contents())
+        replace_elm = elm
+        transform = true
+      }
+      // figurable to be captioned is in the paragraph above
+      else {
```
NGPixel, line 36: `} else {` should be a single line, otherwise eslint will complain.

despens:
Thank you for these suggestions, I will work on them!

Just one question:

> Don't use .caption CSS class name as it's already used by Vuetify.

I tried different options and thought using `.caption` is good _because_ it is included in the default css. For users that have the figCaption module disabled, the caption would still render as a caption, just missing the semantic markup and layout options. My idea was that a caption should look consistent across the app, and opted for the underlying design framework default (still have to integrate that with the `<figcaption>` tag).

Would you still prefer a new css class to be introduced?

sbonaime:
@despens Does this PR could be adapted / improve / use for video captions (like youtube) ?

despens:
@sbonaime These captions are using HTML-native `<figure>` and `<figcaption>`, currently working with `` elements. It is already prepared to work with other elements, which could be `<iframe>`, which is how most videos are embedded, or `<video>`, `<audio>`, etc.

If by "captions" you mean subtitles, or "close captions" however, that's something entirely different and not in scope here.

NGPixel:

despens:
Does this mean I have to submit the pull request again?

NGPixel:
If you can edit this one and just change the target branch, I can reopen it. If it's not possible, then yes you will need to submit a new one.

Issue #1186, opened 2019-11-06
Page Properties should allow switching template type.


arieljake:
Not submitting a feature request...creating the issue to prompt a discussion first. If it's possible and warranted I'll submit on the wiki.

Can Page Properties offer a path to change from a Markdown page to a Code page?

NGPixel:
It will be possible for some editors yes. For example, Markdown to Code or Visual Editor will be possible.

albertosoho:
Hi, is this possible? Can i switch from markdown to html?

NGPixel:
@albertosoho https://requarks.canny.io/wiki/p/switching-from-an-editor-to-another-on-an-existing-page

regevbr:
@NGPixel I might be inclined to implement this feature :-)
I think we can start with switching between visual and html editor, and keep the more "complicated" switching to later on.
Do you have any pointers before I start or should I just go freestyle?

NGPixel:
It should be an option in the Page Properties dialog, which calls a GraphQL resolver (e.g. pages > changeEditor) which calls a method in the pages model. Then reload the page on success, which would load the new editor.

justein230:
This feature would be hugely helpful for my organization, and would allow me to put it into production! @regevbr I watched the GIF and the fix looks fantastic! Excited to see this in prod soon 😃

Ximmer00:
+1 On this implementation

PR #1935, opened 2020-05-25
feat: Page Properties should allow switching template type #1186

regevbr:
fix: #1186

Implement editor changing after page creation.
The implementation ~is pretty straight forward and~ works all ways.
@NGPixel I'm not a UI/UX man so please let me know what changes you want to make.
Also this is my first real time working with vue so excuse my beginners mistakes if such exists

### DIFF
--- a/client/components/editor.vue
+++ b/client/components/editor.vue
@@ -46,7 +46,7 @@
         v-divider.ml-3(vertical)
     v-content
       component(:is='currentEditor', :save='save')
-      editor-modal-properties(v-model='dialogProps')
+      editor-modal-properties(v-model='dialogProps' v-on:editor_changed="saveAndRefresh")
       editor-modal-editorselect(v-model='dialogEditorSelector')
       editor-modal-unsaved(v-model='dialogUnsaved', @discard='exitGo')
       component(:is='activeModal')
@@ -506,6 +506,10 @@ export default {
       } catch (err) {
         // Error is already handled
       }
+    },
+    async saveAndRefresh() {
+      await this.save({ rethrow: true })
+      window.location.reload()
     },
     async exit() {
       if (this.isDirty) {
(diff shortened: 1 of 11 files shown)


regevbr:
wiki

regevbr:
My only drawback is the file extension. Do we need to change it when the editor is changed?
If we don't change it, it might affect auto importing from remote sources (like git)

regevbr:
Not renaming the file can cause issues when moving pages around, as the storage modules calculate the path using the file name extension which can be changed.
I need to fire the renamed event to the storage module when the content type of a page is changed (markdown -> code for example).
Also need to modify the storage modules `renamed` event handle to be able to get the current and new file extension, as the current implementation will fail.
@NGPixel thoughts?

regevbr:
I added code to handle the renaming of the file, works well. The PR is ready for CR

NGPixel:
There's also the matter of converting markdown to HTML and vice-versa when switching editors.

For the renaming situation, I think a new event type altogether would be more appropriate. It may call the standard rename under the hood, but that would allow for more flexibility in storage modules instead of trying to make `rename` handle everything.

regevbr:
@NGPixel thanks for the comments.

About the conversion:
1. From html to markdown we can use https://github.com/domchristie/turndown
2. From markdown to html - can we simply use the rendered html? are there any drawbacks to using it?

About the renaming event:
If you take a look at the code you will see it is not adding any complexity...  The method already has a source/target for most properties, so I just added the extensionType to the mix. Making a new event will be a simple pass through function and the same changes will be required for the existing renamed methods. thoughts?

regevbr:
I implement the type conversion - see gif below.

wiki3

regevbr:
@NGPixel did you have a chance to look at this PR?

NGPixel:

PR #1955, opened 2020-05-28
Open redirect vulnerability mitigation

daneallen:
Registered early middleware function in the express router to explicitly strip repeating / characters found within the req.url variable that  could otherwise be used as an open redirect.

### DIFF
--- a/client/components/admin/admin-security.vue
+++ b/client/components/admin/admin-security.vue
@@ -22,6 +22,15 @@
                 v-card-text
                   v-switch.mt-3(
                     inset
+                    label='Block Open Redirect'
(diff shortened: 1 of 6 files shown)


NGPixel:
This should be part of the security middleware:
https://github.com/Requarks/wiki/blob/master/server/middlewares/security.js

And it should be an optional setting, as this alter the URL and could break a future module (especially auth) / functionality that relies on URL params.

No query param is ever used for redirection in the code anyway. But I'm not against providing the option.

daneallen:
Copy. I'll tackle the correction as you have requested and close this PR.

Issue #1352, opened 2020-01-03
Page rendering stripped up <style> tag 


Jaeha90:
**Describe the bug**
When including a <style> tag in a  page edited with markdown, the resulting HTML page stripped up the <style> tag.

Here is the rendering during the page editing (Which is totally fine) :
![image](https://user-images.githubusercontent.com/7950262/71720891-7e8b1a80-2e23-11ea-9f25-25fe69bf040a.png)

And here is the rendering on the page in view mode :
![image](https://user-images.githubusercontent.com/7950262/71720973-babe7b00-2e23-11ea-8825-df951418508a.png)

When I inspect the html, I can see the <style> tag has been stripped up from the page.
I have investigate a little bit more, and if I check the database for this page, I can see the <style> tag is already stripped up from the render column in pages table :
![image](https://user-images.githubusercontent.com/7950262/71721287-cb232580-2e24-11ea-9ae1-0238e45207d2.png)

Am I missing something ? Knowing that Html is allowed in Markdown rendering configuration.

Thank you

mielipuolinen:
Run into this issue as well. Here's my report about it too.

**Describe the bug**
HTML style-tags are being stripped from Markdown pages which makes page customization more difficult.

**To Reproduce**
Steps to reproduce the behavior:
1. Create an empty page with markdown editor
2. Add <style></style> tags into page with some CSS content.
3. Save and view page source to confirm that style tags are stripped

**Expected behavior**
Style-tags not to be stripped.

**Screenshots**

**Host Info (please complete the following information):**
 - OS: Ubuntu 18.04, up-to-date unmodified docker compose file
 - Wiki.js version: 2.0.12
 - Database engine: PostgreSQL 11.6

**Additional context**
Same behavior is with script-tags which could be useful too for interactive pages (e.g. complex tables). 
Obviously for script-tags you would want to trust your users before allowing them to use those.

mielipuolinen:
@NGPixel Could you take a look at this if possible? Noticed this one was missing an assignee. 
Cheers!

Smankusors:
there's warning on Vue if you attempt to use `style` tag inside page.

```
vue.esm.js:629 [Vue warn]: Error compiling template:

Templates should only be responsible for mapping the state to the UI. Avoid placing tags with side-effects in your templates, such as <style>, as they will not be parsed.
```

so.... this is become more complicated than I think 😕

regevbr:
[...]

PR #1967, opened 2020-05-31
fix: Page rendering stripped up <style> tag #1352

regevbr:
fix: #1352

### DIFF
--- a/server/modules/rendering/html-core/renderer.js
+++ b/server/modules/rendering/html-core/renderer.js
@@ -250,20 +250,24 @@ module.exports = {
     iterateMustacheNode($.root())
 
     // --------------------------------
+    // Escape style elements
+    // --------------------------------
+
+    $('style').each(function(i, item) {
+      item.tagName = 'component';
+      $(this).attr('is', 'style');
+    })
+
+    // --------------------------------
     // STEP: POST
     // --------------------------------
 
-    let output = decodeEscape($.html('body').replace('<body>', '').replace('</body>', ''))
+    let output = decodeEscape($.html('body').replace(/^<body>/, '').replace(/<\/body>$/, ''))
 
     for (let child of _.sortBy(_.filter(this.children, ['step', 'post']), ['order'])) {
       const renderer = require(`../${_.kebabCase(child.key)}/renderer.js`)
       output = await renderer.init(output, child.config)
     }
-
-    // --------------------------------
-    // Escape style elemtns
-    // --------------------------------
-    output = output.replace(/<style/g, `<component is="style"`).replace(/<\/style>/g, `</component>`);
 
     return output
   }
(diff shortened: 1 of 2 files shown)


regevbr:
Adding <style> to templates in vue is not allowed and results in the following warning
```html
Templates should only be responsible for mapping the state to the UI. Avoid placing tags with side-effects in your templates, such as <style>, as they will not be parsed.
```
And the removal of the style from the resulting html.

I managed to find a solution [here](https://stackoverflow.com/a/57331310/2242402):
in the html render we need to replace the style tags with component:
```javascript
    output = output.replace(/<style/g, `<component is="style"`);
    output = output.replace(/<\/style>/g, `</component>`);
```

regevbr:
I actually need to use a dom manipulation so I won't change free text by accident

regevbr:
I used cheerio to safely replace the style elements with component ones.
In order for it to work though, I had to add the `component` element as a safe one to the html-security module.
In order to prevent a misuse of the `component` I added a validation hook that will only allow `component` when it is used as a style element.

The following example:
```markdown
# H1
<style >
.sdf {
   background-color: black;
}
</style>

<component is="h2">
.sdf {
   background-color: black;
}
</component>

<component>
.sdf {
   background-color: black;
}
</component>

sdfsdf
{ .sdf }
```

Will safely compile to:
```html
<h1 class="toc-header" id="h1"><a href="#h1" class="toc-anchor">¶</a> H1</h1>
<component is="style">
.sdf {
   background-color: black;
}
</component>
<p class="sdf">sdfsdf</p>
```

Which will be rendered by vue as:
```html
<div><h1 id="h1" class="toc-header"><a href="#h1" class="toc-anchor">¶</a> H1</h1>
 <style>
.sdf {
   background-color: black;
}
</style>
 <p class="sdf">sdfsdf</p></div>
```

regevbr:
@NGPixel can you please CR?

regevbr:
@NGPixel while you are merging simple PRs would you mind handling this one as well 🙏?

NGPixel:
I see what you're trying to do, but the plan is to have 2 new permission scopes to add scripts and styles from the Page Properties dialog. They will live in the `extra` json field of the page and be rendered outside of the vue instance.

regevbr:
Ok that makes sense.
Until the plan will come together, you don't want to enable custom css per page like I implemented it? If so I will close the PR.
I think we can start by allowing my PR and when the big plan is implemented, we can disable it and force the css to be injected in the page properties...
Thoughts?

regevbr:
@NGPixel I see that you implementer the page js and css features, awesome!
I didn't see that you added the permissions to the admin page yet am I right?

NGPixel:
They were added here: https://github.com/Requarks/wiki/blob/83b83a7510007a58d66aeba0c556ca28151712ee/client/components/admin/admin-groups-edit-permissions.vue#L83-L95

regevbr:
Oh my bad... Awesome work!

PR #1991, opened 2020-06-04
Change reconnectLink behavior for page move

### LINKED ISSUE #1990: Moving pages and links consistency
**Describe the bug**
Moving a page doesn't consistently validate the link URLs. It means moving a page will lead to inconsistency in the validity of URLs (which is shown in blue/red colors)

**To Reproduce**
1. Create page '/test/a'
2. Create page '/test/b' with content `[a](/test/a) and [a new](/test/an)`
3. Everything is OK. the first link is blue and the second one is red.
4. Move '/test/a' to '/test/an'
5. The first link is still blue and the second one is still red. In other words, the blue link is not working but the red one is working.

**Expected behavior**
In the 5th step, we expected to see the first link in red and the second one in blue.

**Screenshots**
image

**Host Info**
 - OS: Docker
 - Wiki.js version: 2.3.81
 - Database engine: PostgreSQL 11.7

sesajad:
Move behavior of `reconnectLinks` was replacing
```
<a href="old-path" class="invalid">
```
with 
```
<a href="new-path" class="valid">
```
but we know that before the move we have valid links like
```
<a href="old-path" class="valid">
```

Another necessary change was that if a page is moved from `old-path` to `new-path` we have to do something more than this. We may have invalid links to the new path, like
```
<a href="new-path" class="invalid">
```
which they shall be valid now.
And this is the same process that is done in page creation.

This fixes #1990

### DIFF
--- a/server/models/pages.js
+++ b/server/models/pages.js
@@ -482,13 +482,20 @@ module.exports = class Page extends Model {
       })
     }
 
-    // -> Reconnect Links
+    // -> Reconnect Links : Changing old links to the new path
     await WIKI.models.pages.reconnectLinks({
       sourceLocale: page.localeCode,
       sourcePath: page.path,
       locale: opts.destinationLocale,
       path: opts.destinationPath,
       mode: 'move'
+    })
+   
+    // -> Reconnect Links : Validate invalid links to the new path
+    await WIKI.models.pages.reconnectLinks({
+      locale: opts.destinationLocale,
+      path: opts.destinationPath,
+      mode: 'create'
     })
   }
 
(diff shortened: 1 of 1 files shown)


server/models/pages.js
```diff
@@ -482,14 +482,21 @@ module.exports = class Page extends Model {
...
       locale: opts.destinationLocale,
       path: opts.destinationPath,
       mode: 'move'
     })
+   
+    // -> Reconnect Links : Validate invalid links to the new path
+    await WIKI.models.pages.reconnectLinks({
+      locale: opts.destinationLocale,
+      path: opts.destinationPath,
+      mode: 'create'
```
NGPixel, line 498: Shouldn't it be `move`?

Issue #2011, opened 2020-06-07
bug: secure html module removes target attribute from links


regevbr:
**Describe the bug**
Setting in the admin html-core rendering options to open external links in new tab doesn't work.
It seems that the target tag is being stripped by the html-security rendering module.

**To Reproduce**
1. Enable the html-security rendering module and enable the sanitize html setting
2. In the html-core rendering module enable the external links to open in new tab setting
3. Add an external link in a page
4. The link has the `rel` tag and the `is-external-link` class, but not the `target` attribute

**Expected behavior**
External links should open in a new window

**Host Info (please complete the following information):**
 - OS: Docker
 - Wiki.js version: 2.4.105
 - Database engine: aws aurora postgress

PR #2012, opened 2020-06-07
fix: secure html module removes target attribute from links #2011

regevbr:
Fix: #2011

Reproduced the bug on master, and verified the fix


regevbr:
@NGPixel can you please CR this? The shortest PR you have ever seen :-)

PR #2018, opened 2020-06-08
oauth2: enabled module, parse user profile from jwt

vketsenko:
Since OAuth2 is not implemented yet and is expected in 2.5, I made simple implementation that works with JWT and loads user profile from JWT data

### DIFF
--- a/server/modules/authentication/oauth2/authentication.js
+++ b/server/modules/authentication/oauth2/authentication.js
@@ -4,24 +4,27 @@
 // OAuth2 Account
 // ------------------------------------
 
+const jwtParser = require('jsonwebtoken');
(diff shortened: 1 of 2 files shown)


NGPixel:
You're assuming that everyone is using JWT on their OAuth2 provider which is not the case.

vketsenko:
I don't assume, I've chosen JWT as fastest option to implement on both sides. Others may add their implementation if JWT is not enough or wait for 2.5

NGPixel:
Except you've modified the Generic OAuth2 module, which is supposed to be "generic" and allow any standard OAuth2 configuration. You can't just modify an existing module to your own specs and call it a day.

So either make your PR compatible with any standard OAuth2 (with an option for JWT parsing) or close it and wait for 2.5.

Issue #1998, opened 2020-06-05
Any plan to support Backlink or/and Graph view?


yrong:
**Question**

Any plan to support Backlink or/and Graph view like [these products](https://www.notion.so/Artificial-Brain-Networked-notebook-a131b468fc6f43218fb8105430304709) ?

Smankusors:
this should be feature request instead

yrong:
after some time play with wiki code base seems backlink is doable from data model layer perspective . The pull request as following has nothing to do with UI side and is more like a experiment than a formal feature commit. 

https://github.com/Requarks/wiki/pull/2023

Hope we can improve back link as first class citizen from all perspective(like tag)

PR #2023, opened 2020-06-09
back link support

yrong:


yrong:
backgroud: https://github.com/Requarks/wiki/issues/1998

server/models/pages.js
```diff
@@ -755,10 +768,18 @@ module.exports = class Page extends Model {
         //   }
         // })
         .first()
+      if (page.backLinks && page.backLinks.length) {
+        page.backLinkPages = []
+        for (backLink of page.backLinks) {
+          backLinkPage = await backLink.$relatedQuery('page').column(['id', 'path', 'title'])
```
regevbr, line 774: This can be optimized with Promise.all, no need to wait page by page

regevbr:
@yrong if I understand correctly, you want to provide links to other locals of the current page?
If so, I think it will make more sense to add the locale to the view... Also, you should probably filter the current page from the back links

yrong:
@regevbr  some fix per your suggestion

https://github.com/Requarks/wiki/pull/2023/commits/a5420095f14309544eba2d5f22d0cea2cc7a6d01

NGPixel:

Issue #1540, opened 2020-02-28
tokenRenewal seems to be hard coded


Trockyyy:
**Describe the bug**
Changing the parameter tokenRenewal on the app (Admin >Authentication) updates well in the database, but still displays "14d" (as default).

Plus, the parameter is not actually used, as it seems to be hard coded in _server/core/auth.js_
`moment().subtract(14, 'days')`

**Host Info:**
 - Wiki.js version: 2.0.12

PR #2042, opened 2020-06-10
fix: tokenRenewal seems to be hard coded #1540

regevbr:
fix: #1540

### DIFF
--- a/server/core/auth.js
+++ b/server/core/auth.js
@@ -2,6 +2,7 @@ const passport = require('passport')
 const passportJWT = require('passport-jwt')
 const _ = require('lodash')
 const jwt = require('jsonwebtoken')
+const ms = require('ms')
 const moment = require('moment')
 const Promise = require('bluebird')
 const crypto = Promise.promisifyAll(require('crypto'))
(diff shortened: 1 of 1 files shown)


regevbr:
@NGPixel can you please CR? Super short fix...

regevbr:
@NGPixel can you please CR? Super short fix...

server/core/auth.js
```diff
@@ -112,7 +113,8 @@ module.exports = {
       if (err) { return next() }
 
       // Expired but still valid within N days, just renew
-      if (info instanceof Error && info.name === 'TokenExpiredError' && moment().subtract(14, 'days').isBefore(info.expiredAt)) {
+      if (info instanceof Error && info.name === 'TokenExpiredError' &&
+        moment.subtract(ms(WIKI.config.auth.tokenRenewal), 'ms').isBefore(info.expiredAt)) {
```
NGPixel, line 117: Should be `moment()`, otherwise you're relying on an existing instance which could have been modified previously.

regevbr: Ops my bad

regevbr: @NGPixel fixed, please review

Issue #1780, opened 2020-04-25
Edit / Page Create Buttons showing up even if no action is allowed


K-Leon:
First of i love your hard work with wiki.js. It is really impressive!

While setting everything up i noticed that in restricted read only usergroups all site edit /create and mange buttons are showing up.

![image](https://user-images.githubusercontent.com/2196713/80291354-1bfcb800-874d-11ea-988f-1979a94b3654.png) 
![image](https://user-images.githubusercontent.com/2196713/80291360-27e87a00-874d-11ea-9400-4f185b533127.png)

In my opinion it is really confusing for the users to see these actions without the ability to do anything about it. The behavior should match anonymous users.

PR #2043, opened 2020-06-10
feat: Edit / Page Create Buttons showing up even if no action is allowed #1780

regevbr:
fix: #1780
I also fixed the view page icon which doesn't exist
I also fixed the page mode when viewing source page

I have removed all buttons (from both menus) if the user doesn't have the necessary global permissions for them

### caveats:
- I removed non relevant buttons, @NGPixel maybe you prefer just to disable them?

### DIFF
--- a/client/themes/default/components/page.vue
+++ b/client/themes/default/components/page.vue
@@ -132,7 +132,7 @@
                   v-spacer
                   v-tooltip(right, v-if='isAuthenticated')
                     template(v-slot:activator='{ on }')
-                      v-btn.btn-animate-edit(icon, :href='"/h/" + locale + "/" + path', v-on='on', x-small)
+                      v-btn.btn-animate-edit(icon, :href='"/h/" + locale + "/" + path', v-on='on', x-small, v-if="hasReadHistoryPermission")
                         v-icon(color='indigo', dense) mdi-history
                     span {{$t('common:header.history')}}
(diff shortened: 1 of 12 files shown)


regevbr:
So I managed to get it working perfectly, by calculating the effective permission on each view you are on, so depending on the page you are at, you will only see actions that you are allowed to do.

@NGPixel can you please CR this? please let me know if there is anything I can do better here

regevbr:
@NGPixel I know this is somewhat a big PR but this should improve the user experience for all the users of wiki.js ...

client/themes/default/components/page.vue
```diff
@@ -491,8 +502,8 @@ export default {
     this.$store.set('page/tags', this.tags)
     this.$store.set('page/title', this.title)
     this.$store.set('page/updatedAt', this.updatedAt)
-    if (this.commentsPermissions) {
-      this.$store.set('page/commentsPermissions', JSON.parse(atob(this.commentsPermissions)))
+    if (this.effectivePermissions) {
+      this.$store.set('page/effectivePermissions', JSON.parse(atob(this.effectivePermissions)))
```
NGPixel, line 506: Even though it wouldn't break in this case, avoid using atob() since it doesn't handle non-latin characters. Instead use:
```js
JSON.parse(Buffer.from(this.effectivePermissions, 'base64').toString())
```

regevbr:
@NGPixel I fixed the json parsing in all places...

Issue #2020, opened 2020-06-08
Caching kroki svgs


regevbr:
When using kroki to generate umls in markdown, every page loads calls the kroki api which is redundant and time consuming. 
@NGPixel do you think that inlining the resulting svg on render will give better performance?
If the answer is yes, I'm willing to create a pr!

twoi:
Two questions about this 
> When using kroki to generate umls in markdown, every page loads calls the kroki api which is redundant and time consuming.

But is this really true?

twoi:
@regevbr 

Your feature is useful for me for a different reason however: When plantuml / kroki rendering is requested by the server, rather than the client, I don't need to expose plantuml / kroki to the internet. That way I don't need to figure out how to do authentication for those.

regevbr:
@twoi you are correct when you look at it from a single client point of view, but there is no need to slow down the first page view for every client who access the page and to create that load on the rendering service...
About authentication, indeed that is also an added bonus

twoi:
The only problem is that in contrast to rendering, while editing a plantuml diagram on a page, a hard-coded plantuml server will be used, leaking potentially private data to https://plantuml.requarks.io (perhaps that was the whole point of wiki.js ;-)

Described here: https://github.com/Requarks/wiki/discussions/3583

So once this is fixed and the configured plantuml server gets used for editing as well (in that case presumably called from the browser), I would need to expose the server on the internet...

regevbr:
Seems like you need to express your valid concerns on your mentioned discussion wo they can address it in the pr there

PR #2047, opened 2020-06-11
feat: Caching kroki svgs #2020

regevbr:
fix: #2020

When using krok/plantumli to generate umls in markdown, every page loads calls the kroki api to get the image, which is redundant, time consuming and imposes load on those free services for no reason.. 
I added a new html renderer that will pre fetch the image and inline it into the html.
The reason it is not being handled by the kroki renderer is that markdown-it plugins must be [sync](https://github.com/markdown-it/markdown-it/blob/master/docs/development.md#i-need-async-rule-how-to-do-it)


regevbr:
@NGPixel can you please CR?

regevbr:
Added support for all types of images, and not just svgs by replacing the url with the renderd image base 64 encoded version

NGPixel:
Just a nitpicking change: replace `pre-fetch` with `prefetch` everywhere (same with `preFetch` to `prefetch`), it's one word.

regevbr:
Sure thing

server/modules/rendering/html-image-prefetch/renderer.js
```diff
@@ -0,0 +1,23 @@
...
+  })
+  const contentType = response.headers[`content-type`]
+  const image = Buffer.from(response.body).toString('base64')
+  element.attr('src', `data:${contentType};base64,${image}`)
+  element.removeClass('prefetch-candidate');
+}
+
+module.exports = {
+  async init($) {
+    const promises = $('img.prefetch-candidate').map((_, element) => {
```
NGPixel, line 18: To avoid confusion, don't use `_` as a variable name unless it's lodash.

server/modules/rendering/html-image-prefetch/renderer.js
```diff
@@ -0,0 +1,23 @@
+const request = require('request-promise')
+
+const prefetch = async (element) => {
```
NGPixel, line 3: You should handle request failures gracefully. If the server doesn't respond or return an error.

regevbr: fixed

server/modules/rendering/html-image-prefetch/renderer.js
regevbr: fixed

regevbr:
@NGPixel please let me know if there is anything else

regevbr:
Thanks for the review and merge!

PR #2058, opened 2020-06-14
change language in edit, history and source pages

### LINKED ISSUE #1578: Markdown edit page bug for RTL
### Actual behavior
edit section on the the left side of the screen and the preview section moved to the right.
text is also aligned to the left
image

### Expected behavior
After 12.13.1 update, the edit section moved to the right side of the screen and the preview section moved to the left. and the text on bouth sections was aligned to the right, exactly as it should for RTL users.
The problem is that suddenly they swapped places as they were prior to the update.

I think that the action that triggered it was restarting of the server. but i'm not 100% sure.
another possible reason is adding my company logo but it sound very strange.
i couldn't find a way to undo the company logo addition. the only thing i managed to do is to remove the logo I've added but it didn't retrieved wiki.js logo.

### LINKED ISSUE #2004: text alignment in markdown editor for RTL pages
**Describe the bug**
When editing a page using the Markdown editor for RTL locale pages (like Hebrew), the page is align to left instead of being aligned to right.
I've already reported this bug (#1578) but it must have been missed between all the other bugs.
The problem was fixed before but returned on one of the previous updates.

**To Reproduce**
To reproduce this bug you just need to create a new page for Hebrew local (or any other RTL I presume) and watch.

**Expected behavior**
The edit section should be on the **right side** and the text written in it should be aligned to the **right**.
The preview section should be on the **left side** and the text written in it should be aligned to the **right**.

**Screenshots**

**Host Info**
 - OS: Windows_NT (win32) 10.0.17763 x64
 - Wiki.js version: 2.4.105
 - Database engine: PostgreSQL 12.1, compiled by Visual C++ build 1914, 64-bit

sesajad:
I guess this change will fix #1578, #2004 and maybe more.

How did I come up with this fix?
1. We had a line

https://github.com/Requarks/wiki/blob/a0618ee4f6b47db53163263e9c57528063508f8e/client/components/editor/editor-markdown.vue#L661

2. At the browser console **for edit pages**, the `siteConfig` was equal to 
```
siteConfig
Object { title: ..., lang: "fa", rtl: false, ... }
```
and it `rtl: true` for **wiki pages**.

3. At the server-side, we use following logic to set locale and rtl. (for all both edit and wiki pages)
https://github.com/Requarks/wiki/blob/122235504654881b6dd78807ded1387f4ea17c16/server/controllers/common.js#L394-L395 
we see that `rtl` parameter uses the localization engine which is i18next.

4. The point is that there was 
https://github.com/Requarks/wiki/blob/122235504654881b6dd78807ded1387f4ea17c16/server/controllers/common.js#L366
**only for wiki pages**, this line would change the language for i18next, therefore, `rtl` value became valid.

5. therefore adding that line to other functions will solve localization problems there.

6. But maybe it's not the best practice. there is a LanguageDetector middleware inside the i18neMW that can detect the language from the path. if the default detector cannot work we can implement our custom LanguageDetector. (more on [here](https://github.com/i18next/i18next-express-middleware#language-detection))

Bonus fact: [i18neMW](https://github.com/i18next/i18next-express-middleware) that we are using is deprecated and its successor is [i18nhMW](https://github.com/i18next/i18next-http-middleware).

### DIFF
--- a/server/controllers/common.js
+++ b/server/controllers/common.js
@@ -44,6 +44,8 @@ router.get(['/a', '/a/*'], (req, res, next) => {
 router.get(['/d', '/d/*'], async (req, res, next) => {
   const pageArgs = pageHelper.parsePath(req.path, { stripExt: true })
 
+  req.i18n.changeLanguage(pageArgs.locale)
+  
   const versionId = (req.query.v) ? _.toSafeInteger(req.query.v) : 0
 
   const page = await WIKI.models.pages.getPageFromDb({
@@ -91,6 +93,8 @@ router.get(['/e', '/e/*'], async (req, res, next) => {
     return res.redirect(`/e/${pageArgs.locale}/${pageArgs.path}`)
   }
 
+  req.i18n.changeLanguage(pageArgs.locale)
+  
   // -> Set Editor Lang
   _.set(res, 'locals.siteConfig.lang', pageArgs.locale)
   _.set(res, 'locals.siteConfig.rtl', req.i18n.dir() === 'rtl')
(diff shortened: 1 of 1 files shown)


talalvas:
Thank you very much. When will it going to be implemented?

sesajad:
@talalvas 
You're welcome. I really don't know, It's now pending for review. I guess it will be available in the first release after the merge!

sesajad:
@NGPixel could you please take a look at this?

talalvas:
Any idea if it will be implemented in the next release?
Our staff is struggling with the current layout for some reason...
Thank you very much guys!
Your work is extremely appreciated!

NGPixel:

PR #2070, opened 2020-06-16
Support for client side debugging in VSCode

grbd:
Hi,
I've added some changes to allow for debugging the client side of the code via VSCode
so you should be able to add breakpoints to files within the client directory in vscode and have them hit when the browser reaches that point

I was looking into this as I'm interested in writing my own theme

  * code-workspace file - just an easy way / shortcut of double clicking to open up in vscode, without doing File Open folder.
  * changes to .babelrc / dev/webpack/webpack.dev.js - enable source maps to allow the browser to map back to the code in vscode
  * changes to extensions.json / launch.json - allow for launching of firefox / server or both

I did try to get this working with the chrome debugger for vscode initially but I think it has some compatibility issues with the generated source maps. It breakpoints on regular .js files, but then gets confused which line to map to for the code inside .vue files

At a guess it might be babel I'm not sure.
I have other projects with vue files that don't have this problem but they don't use babel so that's just a guess.
I was able to get around it by using the firefox debugger instead which seems to handle it without any problems

The only downside is that you have to refresh the page after launching to trigger any immediate breakpoints, since it takes a few seconds for firefox to recognise the set breakpoints.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,11 @@
 {
   "comments": true,
+  "env": {
+    "development": {
+      "sourceMaps": true,
+      "retainLines": true
+    }
+  },
   "plugins": [
     "lodash",
     "graphql-tag",
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -4,6 +4,7 @@
     "dbaeumer.vscode-eslint",
     "christian-kohler.path-intellisense",
     "mrmlnc.vscode-puglint",
-    "octref.vetur"
+    "octref.vetur",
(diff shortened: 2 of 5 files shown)


NGPixel:
Remove the code-workspace file. This file can end up being completely different per developer and not really useful for only 1 project by default.

grbd:
ok,
let me know if you want me to remove "firefox-devtools.vscode-firefox-debug" from extensions
I wasn't sure if to add that since so far it's only the firefox debugger I've managed to get working fully with the source maps

NGPixel:

Issue #2072, opened 2020-06-17
bug: no markdown support for underline


regevbr:
There is currently no markdown way to add an underline.
The solution is to use html syntax
```html
<u>underline</u>
```
I think we should add support for underline parsing. The issue is that underline is usually achieved by `__underline__` but markdown-it treats it as bold...
We need to think of another way to declare underline (maybe `____`) and add a plugin for it
@NGPixel what do you think?

PR #2073, opened 2020-06-17
fix: no markdown support for underline #2072

regevbr:
fix: #2072 
Add support for underline in markdown by using `++underline++`.
This fix will also show up in the editor


NGPixel:
2 things to improve:

- Make it optional by adding a toggle as part of the Markdown Core module. Since this is not compliant with the CommonMark specification, it shouldn't be mandatory.

- Avoid adding a new npm dependency for just [10 lines of code](https://github.com/arve0/markdown-it-underline/blob/master/index.js). Include the code directly.

regevbr:
Thanks @NGPixel for the comments.

I didn't use the library I mentioned in the issue as it will conflict with the current behavior of `_sdf_` so I used a different one - `markdown-it-plugin-underline` which I can't locate its repository but it is a bit more complicated and contains 120 lines of code. So I don't think in-lining it will make sense.

About the toggle, I will add it

regevbr:
Done, can you please review?

NGPixel:
Just not a fan of including an npm module with no public source available.

I think it would be more convenient and intuitive to use the `_word_` syntax, which the other module does.

regevbr:
Ok, but itwill collide with current behavior causing change of ui for existing pages. If you are ok with it, I will be happy to do it

NGPixel:
That's fine. It's an optional feature that can be turned off and almost everyone is using the `*word*` syntax for italic already. `_word_` is simply more intuitive as underline for everybody.

regevbr:
@NGPixel I fixed it like you suggested, but I found myself duplicating the plugin code in both server and client. Is there a better way for it? (using the npm package is one way but you said you prefer not to)

NGPixel:
Yeah it's not ideal for now, but the goal is to use the server pipeline in the future, eliminating the need to duplicate everything on the client.

regevbr:
Thanks for the review and merge!

Issue #2074, opened 2020-06-17
bug: playing a video from assets is super slow


regevbr:
**Describe the bug**
When using `<video>` to add a video player that will play a video from an uploaded asset works super slow. I believe the issue is because the server doesn't support chunks.

We need to support chunked requests - see [example](https://github.com/daspinola/video-stream-sample/blob/master/server.js)

I will work on a PR

regevbr:
In general we need to stream the assets instead of reading them as a whole and sending them...

PR #2087, opened 2020-06-19
bug: playing a video from assets is super slow #2074

regevbr:
fix: #2074

Backstory - I uploaded a large 100mb mp4 video as an asset (using postgres as the db) and added an html5 video element to play it in a page. As soon as the page was trying to load, my wiki server just stopped responding for a good few minutes. Cpu was spiking at 100%.
After a lot of investigation, I came to realize that the issue occurs due to a performance bug in the `node-postgress` library. I already fixed [the bug](https://github.com/brianc/node-postgres/pull/2241) ~~and waiting for CR on it.~~ and it was released and it is already being used in the master branch.

This PR fixes a performance issue while handling videos:
When a request is aborted (due to change of playing position for example) the browser cancels the request, but the error handling in the code ignored it and tried to send the file again from DB which wastes time and causes more issue and the response is already closed.

In addition, I thought about an optimization - instead of fetching assets from db and caching them, the assets might already be cached on the local disk on one of the storage modules (git or disk), so why not use them?
The reason I came up with the optimization is that even after my fix to node-postgres it still takes 40 whole seconds to retrieve the 100mb video file from the DB and it is simply a bad user experience...

### DIFF
--- a/server/models/assets.js
+++ b/server/models/assets.js
@@ -6,6 +6,7 @@ const path = require('path')
 const fs = require('fs-extra')
 const _ = require('lodash')
 const assetHelper = require('../helpers/asset')
+const Promise = require('bluebird')
 
 /**
  * Users model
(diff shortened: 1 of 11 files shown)


server/models/assets.js
```diff
@@ -6,6 +6,7 @@ const path = require('path')
 const fs = require('fs-extra')
 const _ = require('lodash')
 const assetHelper = require('../helpers/asset')
+const BB = require('bluebird')
```
NGPixel, line 9: Use `const Promise =` for consistency across the project.

regevbr:
Im using it so it won't collide with the built in Promise, but as you wish 😀

NGPixel:
Bluebird implements the standard Promise spec so it's completely fine.

regevbr:
I'm using typescript 99% of the time and the type definitions are a bit different and can cause minor conflicts, that is why I'm used to do what I did. Will fix in 2 hours.
Any other things to change?

server/models/assets.js
regevbr: fixed

regevbr:
@NGPixel the [bug](https://github.com/brianc/node-postgres/pull/2241) in the postgress library is finally fixed and released. Can you please finish the CR on this? I would love to see it in the upcoming 2.5 release.
Thanks!

regevbr:
@NGPixel I rebased the branch so it won't mix commits. Githubs builtin merge with master behavior is not smooth in my opinion, so I do it manually using rebase and force push

regevbr:
@NGPixel I see you already upgraded the `pg` dependency in the master branch so the stuck node process issue is solved. This PR still fixes a bug (re-fetching a resource after the request was cancelled) and introduces a performance enhancement of serving static files from storage models that use the local disk. Can you please push this PR before the release of 2.5?

Issue #1884, opened 2020-05-14
Tags visibility


JamesLipski:
**Describe the bug**
Tags that are created on pages that are only visible to certain groups, show up in the "Browse By Tags" section

**To Reproduce**
- Create a page in a folder with unique tags.
- From Admin >>  Groups >> Select a group that you don't want to let access to that page (such as Guests) >> Page Rules
-- Create a deny rule, with every role checked/ and 'path starts with' the folder name created in the first step.
- While page/folder doesn't show up (and manual access leads to an unauthorized page) searching the "Browse By Tags" reveals those unique tags (granted no page shows up)

**Expected behavior**
It would be great if those tags that are only visible to certain groups, aren't visible to all.

**Host Info (please complete the following information):**
 - OS: Debian 9.12
 - Wiki.js version: 2.3.81
 - Database engine: postgres 9.6

regevbr:
I will work on a pr to fix

PR #2100, opened 2020-06-24
fix: tags are not filtered by access #1884

regevbr:
fixes: #1884

Tags retrieval and search will always return all tags, regardless of what the user is allowed to see.
This PR fixes this issue.

### Caveats
- Tags that are not associated to any pages will never be returned! Not sure what is the right course of action here - should we always return those orphan tags? If we do than that might impose a security leak...
- This makes the queries a bit slower as we need to get the pages first so we can filter by access
- We can't impose the search result size limit in the query itself as we can receive many duplicates, so we have to get all the tags and then extract the first 5 results. I thought about reversing the query to get the tags and join with the pages, but then if we ask for top 5 we might remove some of them due to access restrictions and we will loose other good results.
- There is no specific order to the tags when making a search, maybe we can improve it somehow

### DIFF
--- a/server/graph/resolvers/page.js
+++ b/server/graph/resolvers/page.js
@@ -181,7 +181,8 @@ module.exports = {
           path: r.path,
           locale: r.locale
         })
-      }).flatMap(r => r.tags)
+      })
+        .flatMap(r => r.tags)
       return _.orderBy(_.uniqBy(allTags, 'id'), ['tag'], ['asc'])
     },
     /**
@@ -211,7 +212,8 @@ module.exports = {
           path: r.path,
           locale: r.locale
         })
-      }).flatMap(r => r.tags)
+      })
+        .flatMap(r => r.tags)
         .map(t => t.tag)
       return _.uniq(allTags).slice(0, 5)
     },
--- a/server/graph/resolvers/page.js
+++ b/server/graph/resolvers/page.js
@@ -171,26 +171,49 @@ module.exports = {
      * FETCH TAGS
      */
     async tags (obj, args, context, info) {
-      return WIKI.models.tags.query().orderBy('tag', 'asc')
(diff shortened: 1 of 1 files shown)


regevbr:
@NGPixel can you please review? This fixes an important security issue

NGPixel:
My issue is that tags were never meant to be made private based on whether they are used on a page or not. They are simply "categories".

Let's say you set the tag "foo" on multiple pages. Then you add a page rule to limit access to these pages if they have the tag "foo" to make them private to a specific group.

Now some editor with less permissions (and with no access to these private pages) creates a new page and adds the same tag "foo". He saves and bam, he gets an access denied page and he has no idea why. We could also filter the tag suggestions to remove any tags already used by pages he doesn't have access to. But that doesn't prevent him from entering it manually and he would still get the same error in the end.

Also, like you mentioned, we get into all sorts of querying and filtering issues as a result.

I would rather explicitly mention that tags are public and always accessible in the docs.

regevbr:
@NGPixel your use case is effective before this fix in any case, am I missing something here?

I agree that my fix has downsides, but let's take for example my company's use case where the wiki host many private information and many public information. We use many tags that I don't want to be exposed. If we don't make some sort of fix, I will have to stop using tags all together and have no way of enforcing all my employees not to use tags...

Maybe we can find a middle ground here - the tags page (`/t`) which is open to all authenticated users should be filtered by security access, wheres the page edit tags will show all available tags...

I still believe my fix security wise is more important than the minor performance impacts.
Thoughts?

regevbr:
We can also disallow that user from your use case to use that tag if he doesn't have a permission for it...

regevbr:
@NGPixel you added a lot of security fixes/enhances for the next release, and I think this one should also be a part of it.
How do you want to proceed here?

Issue #2071, opened 2020-06-16
GraphQL error with MySQL and FULL OUTER JOIN


glogrus:
**Describe the bug**
Page "/a/pages/visualize" has the following error:
```
[GraphQL error]: Message: select `pages`.`id` as `id`, `pages`.`path` as `path`, `title`, `pageLinks`.`path` as `link`, `pageLinks`.`localeCode` as `locale` from `pages` full outer join `pageLinks` on `pages`.`id` = `pageLinks`.`pageId` where `pages`.`localeCode` = 'en' - You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'outer join `pageLinks` on `pages`.`id` = `pageLinks`.`pageId` where `pages`.`loc' at line 1, Location: [object Object], Path: pages,links
```

**To Reproduce**
1. Go to '/a/pages'
2. Click on 'Visualize'
3. See error

**Expected behavior**
Dendrogram representation of your pages.

**Host Info (please complete the following information):**
 - OS: CentOS 7
 - Wiki.js version: 2.4.107
 - Database engine: MySQL 8.0.20

**Additional context**
We don't have FULL JOINS on MySQL, but we can sure emulate them.

In file: server/graph/resolvers/pages.js replace code:
```
    /**
     * FETCH PAGE LINKS
     */
    async links (obj, args, context, info) {
      let results = []

      results = await WIKI.models.knex('pages')
        .column({ id: 'pages.id' }, { path: 'pages.path' }, 'title', { link: 'pageLinks.path' }, { locale: 'pageLinks.localeCode' })
        .fullOuterJoin('pageLinks', 'pages.id', 'pageLinks.pageId')
        .where({
          'pages.localeCode': args.locale
        })
```
with:
```
    /**
     * FETCH PAGE LINKS
     */
    async links (obj, args, context, info) {
      let results = []

      results = await WIKI.models.knex('pages')
        .column({ id: 'pages.id' }, { path: 'pages.path' }, 'title', { link: 'pageLinks.path' }, { locale: 'pageLinks.localeCode' })
        .leftOuterJoin('pageLinks', 'pages.id', 'pageLinks.pageId')
        .where({
          'pages.localeCode': args.locale
        }).unionAll(WIKI.models.knex('pages')
        .column({ id: 'pages.id' }, { path: 'pages.path' }, 'title', { link: 'pageLinks.path' }, { locale: 'pageLinks.localeCode' })
[...]

PR #2104, opened 2020-06-24
fix: GraphQL error with MySQL and FULL OUTER JOIN #2071

regevbr:
fix: #2071

full outer join doesn't exist on mysql and sqlite.
right join doesn't exist in sqlite.
The solution is implemented based on https://www.sqlitetutorial.net/sqlite-full-outer-join/

### DIFF
--- a/server/graph/resolvers/page.js
+++ b/server/graph/resolvers/page.js
@@ -246,20 +246,31 @@ module.exports = {
      * FETCH PAGE LINKS
      */
     async links (obj, args, context, info) {
-      const results = await WIKI.models.knex('pages')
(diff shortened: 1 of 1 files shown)


NGPixel:
For best performance, I would suggest keeping the original implementation for postgres and mssql, while falling back on your version for mysql/mariadb/sqlite, as done here for example:

https://github.com/Requarks/wiki/blob/6ef7b0f130423df7cb814fc658360e9fa3bb8f8d/server/models/assetFolders.js#L37-L60

regevbr:
@NGPixel done

Issue #2078, opened 2020-06-17
Delete page with comments fails with error 


ptptakakumum001:
**Describe the bug**
Trying to delete a page fails with error:
`delete from "pages" where "id" = $1 - update or delete on table "pages" violates foreign key constraint "comments_pageid_foreign" on table "comments"`

**To Reproduce**
1. Administrator. Create a new page
2. User. Make a comment
3. Administrator. Try to delete the page

**Expected behavior**
Page should be deleted without error.

**Host Info (please complete the following information):**
 - OS: Debian
 - Wiki.js version: 2.4.105
 - Database engine: PostgreSQL 11.7

ptptakakumum001:
Similar task: https://github.com/Requarks/wiki/issues/1119

Smankusors:
> Similar task: #1119

yeah it's already fixed. Please update your Wiki.js version. 🙂

ptptakakumum001:
> yeah it's already fixed. Please update your Wiki.js version. slightly_smiling_face

What version is fixed? I have a 105 version, and the last one is 107.

NGPixel:
Please write the correct version. You most likely mean 2.4.105. 105 is simply the build number, not the full version.

ptptakakumum001:
> Please write the correct version. You most likely mean 2.4.105. 105 is simply the build number, not the full version.

by package.json
`  "name": "wiki",
  "version": "2.4.105",`

ptptakakumum001:
Guys?

NGPixel:
@ptptakakumum001 Please **READ** this page before posting these kind of comments:
https://docs.requarks.io/releases/about

ptptakakumum001:
> @ptptakakumum001 Please **READ** this page before posting these kind of comments:
> https://docs.requarks.io/releases/about

Sorry, I didn’t know that you were alone.

PR #2156, opened 2020-07-08
Solving problem with foreing key contraint when page have comments #2078

higortavares:


regevbr:
You should undo all the styling changes as they are not a real part of the fix and they don't conform with the code style of the project.

higortavares:
> You should undo all the styling changes as they are not a real part of the fix and they don't conform with the code style of the project.

Ok. this is because my text editor

regevbr:
I'm wondering what happens if there are multiple ids that get deleted... Not sure that is a feasible scenario currently but we should defiantly handle it as well

higortavares:
> I'm wondering what happens if there are multiple ids that get deleted... Not sure that is a feasible scenario currently but we should defiantly handle it as well

I thought about smooth deletion, but that would cause another problem because the page has a unique key constraint

regevbr:
Why not delete all the comments of all the pages being deleted?

NGPixel:

higortavares:
> Change target branch to `dev` for this PR.

ok

PR #2160, opened 2020-07-09
Editor WYSIWYG and Videos: Improving/Fixing "Insert Media"

KingRial:
When using WYSISYG editor with "Insert media" button, the inserted video was not visible on view mode (See: https://github.com/Requarks/wiki/issues/1459)

Modified code to:

- handle issue https://github.com/Requarks/wiki/issues/1459
- support also video URLs from inside wiki.js uploaded assets

**NB** Iframes must be enabled otherwise embedding external videos (Eg. Youtube video) is not possible ^^
image

PR #2182, opened 2020-07-15
feat: VideoJS support

KingRial:
Just sharing some work which could be considered interesting.

To start to allow an improved customization of "<video>" tags, I have added the support for [VideoJS](https://videojs.com/) library.

- The library usage can be enabled/disabled from administration under "mediaplayers" options.
- When enabled, the rendering pipeline will transform all "<video>" tags into VUE custom element "<videojs>"; as a result videoJS will be the player used to handle videos sources

The videojs support is fully compatible with previous video improvements offered by pull request: https://github.com/Requarks/wiki/pull/2160


NGPixel:

Issue #2167, opened 2020-07-11
plantuml preview in markdown editor does not respect server settings 


ProgramFan:
**Describe the bug**
When the plantuml server in admin:rendering:markdown-html:plantuml is set to a custom server, the rendered page uses that server. But the plantuml preview in the markdown editor still uses https://plantuml.requarks.io. This causes the editor to not be able to display plantuml preview if works offline.

**To Reproduce**
1. Create a docker installation of wiki.js with plantuml server.
2. Configure the wiki to use http://localhost:8080 for plantuml server.
3. Disconnect the host from the Internet.
4. Create a plantuml diagram in the markdown editor.
4. See the preview displays "uml_diagram" instead of the preview.

**Host Info (please complete the following information):**
 - OS: Docker
 - Wiki.js version: 2.4.107
 - Database engine: sqlite 3.11

**Additional context**
Use developer tools of firefox on the preview shows that the url of the plantuml diagram points to "https://plantuml.requarks.io" while the url of the diagram in the rendered page points to "http://localhost:8080". The source code of plantuml.js of the markdown editor shows that it does accept a server configuration. So it might be a regession.

NGPixel:
The client renderer is using a fixed configuration and won't reflect the server renderer. This will be addressed in a future update to have the client use the exact same configuration as the server.

ProgramFan:
Thanks for your reply. Does this mean that there is a way to configure the client renderer? How can I configure it?

NGPixel:
No, it's not configurable.

AGrzes:
Is a PR on this welcome or is somebody else working on this enhancement?

NGPixel:
@AGrzes As long as the PR addresses the full rendering pipeline, and not just PlantUML. If you're interested on working on this enhancement, open a draft PR first to discuss the implementation, before writing any code.

AGrzes:
GtiHub is not allowing me to open draft PR without any change in code so I wlll put my proposed changes here.

1. Reproduce code fetching renders configuration from `client/components/admin/admin-rendering.vue` in `client/components/editor.vue`. 
2. Pass renderers configuration as a property to editor component
3. In `client/components/editor/editor-markdown.vue` Wrap `md` configuration in a function that accepts  renderers configuration
4. Create `md` as computed property of markdown editor component that will call the previously defined function to initialize markdown engine based on renderers config property
5. Use `md` computed property instead of `md` module variable.
6. Update `md` initialization to use renderers configuration
[...]

PR #2219, opened 2020-07-22
feat: client side renderers options

AGrzes:
Addresses #2167

# Implementation Plan

1. Reproduce code fetching renders configuration from client/components/admin/admin-rendering.vue in client/components/editor.vue.
2. Pass renderers configuration as a property to editor component
3. In client/components/editor/editor-markdown.vue Wrap md configuration in a function that accepts renderers configuration
4. Create md as computed property of markdown editor component that will call the previously defined function to initialize markdown engine based on renderers config property
5. Use md computed property instead of md module variable.
6. Update md initialization to use renderers configuration

### DIFF
--- a/client/components/editor.vue
+++ b/client/components/editor.vue
@@ -45,7 +45,7 @@
           span.white--text(v-if='$vuetify.breakpoint.lgAndUp') {{ $t('common:actions.close') }}
         v-divider.ml-3(vertical)
     v-main
-      component(:is='currentEditor', :save='save')
+      component(:is='currentEditor', :save='save', :renderers='renderers' v-if='renderers')
       editor-modal-properties(v-model='dialogProps')
       editor-modal-editorselect(v-model='dialogEditorSelector')
       editor-modal-unsaved(v-model='dialogUnsaved', @discard='exitGo')
@@ -62,6 +62,8 @@ import { get, sync } from 'vuex-pathify'
 import { AtomSpinner } from 'epic-spinners'
 import { Base64 } from 'js-base64'
 import { StatusIndicator } from 'vue-status-indicator'
+
+import renderersQuery from 'gql/admin/rendering/rendering-query-renderers.gql'
 
 import editorStore from '../store/editor'
 
@@ -172,7 +174,8 @@ export default {
         title: '',
         css: '',
         js: ''
-      }
+      },
+      renderers: null
     }
   },
   computed: {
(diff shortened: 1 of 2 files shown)


client/components/editor.vue
```diff
@@ -63,6 +63,8 @@ import { AtomSpinner } from 'epic-spinners'
 import { Base64 } from 'js-base64'
 import { StatusIndicator } from 'vue-status-indicator'
 
+import renderersQuery from 'gql/admin/rendering/rendering-query-renderers.gql'
```
AGrzes, line 66: I think it is not right to call admin query here but duplicating the same query in editor section also do not feel right. Any suggestions how to proceed?

client/components/editor.vue
```diff
@@ -63,6 +63,8 @@ import { AtomSpinner } from 'epic-spinners'
 import { Base64 } from 'js-base64'
 import { StatusIndicator } from 'vue-status-indicator'
```
lelaarnett, line 64: Approve

AGrzes:
From my point of wiew this is finished - please let me know if any changes are still required.

NGPixel:
While this work because you are admin, the renderers query will fail for normal users as the `rendering.renderers` resolver requires the `manage:system` permission. Could be changed to write:pages:

https://github.com/Requarks/wiki/blob/95b6a7ad823803b3452169196935426c2a2345f0/server/graph/schemas/rendering.graphql#L21

As for the query itself, inline it directly in editor.vue. External queries are gradually being removed from the project to reduce complexity. e.g.:
```js
import gql from 'graphql-tag'
// ...
query: gql`
{
  rendering {
    ...
  }
}
`
```

AGrzes:
@NGPixel - I added commits handling Your suggestions

AGrzes:
Merget latest changes from `dev` and did small fix.

AGrzes:
Merged latest changes from `dev`

EugenDueck:
@NGPixel I too would like to prevent our internal data from going to some external server, so I'm very interested in this PR. What is missing for it to be merged?

mbedded:
> @NGPixel I too would like to prevent our internal data from going to some external server, so I'm very interested in this PR. What is missing for it to be merged?

Is there a plan when this PR will be merged to the next version of wikijs? It is a great product and I have the same issue that i want to prevent potential data-leaks to external services i can't  control. Do we have to wait until wikijs-vnext is released until this feature is added?

yumetodo:
any updates?

VusalDev:
any updates?

awschult002:
this is a significant security issue for many companies, which is preventing widespread adoption. What can we do to get this PR complete?

agayev169:
@NGPixel is there anything stopping this PR from completing? Is there anything we can be helpful with?
This issue stops companies from using Wiki JS with PlantUML due to security reasons, it would be great if we could complete it.

Issue #1672, opened 2020-04-05
Showing not all search results


Annastasiiya:
### Actual behavior
I cannot see all list search result https://prnt.sc/rtde5z, https://prnt.sc/rtdg8a 
### Expected behavior
I should see all the search results
### Steps to reproduce the behavior
Enter search word in search string.

PR #2232, opened 2020-07-27
Add overflow scroll to search results

puppycodes:
All search results were not visible on desktop screens beyond the viewport height, adding overflow-y: scroll fixes this

ezgif-1-671563cb08f8

### DIFF
--- a/client/components/common/search-results.vue
+++ b/client/components/common/search-results.vue
@@ -166,6 +166,7 @@ export default {
   position: fixed;
   top: 64px;
   left: 0;
+  overflow-y: scroll;
   width: 100%;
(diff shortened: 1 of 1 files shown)


Jhobean:
https://github.com/Requarks/wiki/issues/1672

NGPixel:
Should be `auto`, not `scroll` as otherwise it will always display a scrollbar on larger screen (on non-Mac machines at least).

puppycodes:
> Should be `auto`, not `scroll` as otherwise it will always display a scrollbar on larger screen (on non-Mac machines at least).

updated ;)

PR #2282, opened 2020-08-10
Fix and enable OIDC auth method

dbnicholson:
This gets the OIDC authentication method working properly with current wiki and passport-openidconnect.

### DIFF
--- a/server/modules/authentication/oidc/authentication.js
+++ b/server/modules/authentication/oidc/authentication.js
@@ -19,17 +19,19 @@ module.exports = {
         issuer: conf.issuer,
         userInfoURL: conf.userInfoURL,
         callbackURL: conf.callbackURL
-      }, (iss, sub, profile, jwtClaims, accessToken, refreshToken, params, cb) => {
-        WIKI.models.users.processProfile({
-          id: jwtClaims.sub,
-          provider: 'oidc',
-          email: _.get(jwtClaims, conf.emailClaim),
-          name: _.get(jwtClaims, conf.usernameClaim)
-        }).then((user) => {
-          return cb(null, user) || true
-        }).catch((err) => {
-          return cb(err, null) || true
-        })
+      }, async (iss, sub, profile, cb) => {
+        try {
+          const user = await WIKI.models.users.processProfile({
+            profile: {
+              ...profile,
+              email: _.get(profile, '_json.' + conf.emailClaim)
+            },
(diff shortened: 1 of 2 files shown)


benito103e:
Hi,
I tried the OIDC generic  authentication method with your fixes.
But I experienced a problem when getting authorization token :
2 requests are sent from wikijs to OIDC identity provider : 
- The first one is ok
- The second request get a 400 response with "Authorization code has been revoked" error
Did you already experience the same issue ?

benito103e:
It's ok if I skip user profile : `skipUserProfile: true,`
So I don't have the same behaviour than you and I need to instantiate user with : 

```
const user = await WIKI.models.users.processProfile({
            profile: {
              id: jwtClaims.sub,
              email: _.get(jwtClaims, 'email'),
              name: _.get(jwtClaims, 'name')
            },
            providerKey: 'oidc'
          })
```

dbnicholson:
No that didn't happen for me. Does your provider normally allow requesting the user profile? Does it support the `profile` scope? Is your client doing a `code` flow or an `id_token` flow?

I suppose `skipUserProfile` could be an option and you would attempt to get the info from the id token claims like this.

benito103e:
Yes I think skipUserProfile option is required.
Because the retrieval of user profile data is not part of OIDC core.

And it is one of the possible options in the passport-openidconnect (identifierField, scope, passReqToCallback, skipUserProfile, sessionKey, store, authorizationURL, tokenURL).

We should also keep emailClaim and usernameClaim options to extract them from IdToken.

dbnicholson:
`usernameClaim` to me implies an actual username but wiki uses the email address for a username. I think `nameClaim` would be clearer, but I don't feel that strongly about it.

My reading of the oidc spec is that for the `code` flow that provides an access token, the userinfo endpoint should be accessible - https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims. However, I don't see any reason to keep that option out of you know the id token contains all the necessary claims.

benito103e:
I totally agree with you on naming nameClaim.

you're right about "UserInfo Endpoint" : this is described in core specs but it's frequently considered optional and we should have all data we need in IDToken.

NGPixel:
Thanks 👍

dbnicholson:
@NGPixel thanks. @benito103e I'll try to get a little time this week to open another PR to address the comments we discussed.

PR #2334, opened 2020-08-25
Allow to control logLevel from docker-compose.yml

BuhtigithuB:


NGPixel:
This is causing issues in some installations, which is why it was recently removed.

Also setting it to anything lower than info will not display anything more than info. This will be addressed in 3.0.

PR #2345, opened 2020-08-28
feat: export creation date in dumped content

jaljo:
Hello all !

We're using wiki js as an internal tool, and we enjoy it :slightly_smiling_face:  We're making good use of the storage plugin to backup created articles in a github repository and them use it as a source for our blog. That is, we're missing a critical information, that we believe should be exported when creating pages in the storage: the creation date.

This PR also fixes the jest tests for the `injectPageMetadata` helper, which were broken.

Thank you for your time !


NGPixel:
Just one tiny change:

Use `dateCreated` instead of `date_creation`

PR #2354, opened 2020-09-01
fix: secure html module removes allowfullscreen, allow and frameborder attributes from iframes

Wani4ka:


server/modules/rendering/html-security/renderer.js
```diff
@@ -12,6 +12,9 @@ module.exports = {
 
       if (config.allowIFrames) {
         allowedTags.push('iframe')
+        allowedAttrs.push('allowfullscreen')
```
NGPixel, line 15: allowfullscreen is redundant with allow and is deprecated.

```suggestion

```

server/modules/rendering/html-security/renderer.js
```diff
@@ -12,6 +12,9 @@ module.exports = {
...
+        allowedAttrs.push('allow')
+        allowedAttrs.push('frameborder')
```
NGPixel, line 17: frameborder is deprecated. Use CSS instead.

```suggestion

```

NGPixel (approved):

PR #2373, opened 2020-09-06
fix: migration error for new installs for postgresql

RainshawGao:


NGPixel:
This migration is to re-create the isEnabled field if it isn't present already. So the goal is not to alter the column. The issue was the if() condition which has been fixed. Thanks anyway!

Issue #2091, opened 2020-06-20
overflow content tabs should be scrollable


brietsparks:
**Describe the bug**
Content tabs that overflow beyond parent width are inaccessible.

I think the fix is to set `overflow: scroll` in the [tabset styles](https://github.com/Requarks/wiki/blob/5d43f6ada1d2492a3d67f967560e2a9824845d51/client/themes/default/components/tabset.vue#L67). If this is still unaddressed by the time I am able to set up a dev environment, then I'll make a PR. Otherwise, someone will hopefully make this quick change, verify, and push it.

**To Reproduce**
1. create some content tabs that take up more width than their parent 

**Expected behavior**
A user should be able to access overflowed tabs.

**Screenshots**
<img width="757" alt="Screen Shot 2020-06-20 at 6 04 05 PM" src="https://user-images.githubusercontent.com/5377854/85213206-d4477680-b320-11ea-99cf-fbc9d40bbc00.png">

bernieke:
The same happens when you have a mermaid diagram which is too wide. On Safari it's apparently possible to do a horizontal scroll, but on Chrome this doesn't work.

PR #2442, opened 2020-09-14
fix: tabset tabs can't be scrolled

Wani4ka:
Fix for #2091. The change advised in that issue works brilliantly.


Wani4ka:
@NGPixel any updates on this?

PR #2458, opened 2020-09-16
Add width and height to allowed markdown attributes

noisymime:
Currently if using the markdown files outside of wiki.js (In my case with pandoc) it's impossible to specify the size of an image that works in both places as the width/height attributes are stripped in wiki.js. Whilst it may be possible to do this using a class (Which is allowed), this does not allow for specific per image width/height values. 

Is there any reason that these cannot be included in the allowed attributes?

### DIFF
--- a/server/modules/rendering/markdown-core/renderer.js
+++ b/server/modules/rendering/markdown-core/renderer.js
@@ -40,7 +40,7 @@ module.exports = {
     }
 
     mkdown.use(mdAttrs, {
-      allowedAttributes: ['id', 'class', 'target']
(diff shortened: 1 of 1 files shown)


NGPixel:
Image size can be defined per image using this syntax:
```
![image](/path/to/image.jpg =100x50)
```

You can also omit either value for auto. (e.g. `=100x` or `=x50`)

noisymime:
I understand that this can be done with the syntax mentioned, this is non-standard in markdown and is not recognised by tools such as pandoc. 

What are the downsides to having these included in the whitelist?

PR #2551, opened 2020-10-07
feat: search by date with elasticsearch

dwojcik92:
Added functionality to search by date with elasticsearch. The createdAt attribute is used.
To search just type
```
date:2020-10-01,2020-10-06
```
for search with range or
```
date:2020-10-01
```
for search with the exact date.

### DIFF
--- a/server/modules/search/elasticsearch/engine.js
+++ b/server/modules/search/elasticsearch/engine.js
@@ -86,65 +86,93 @@ module.exports = {
   async query(q, opts) {
     if(q.search("date:")==0){
       // get the date range
-      
-      try {
-        const results = await this.client.search({
-          index: this.config.indexName,
-          body: {
-            query: {
-              bool: {
-                filter: [
-                  {
-                    bool: {
-                      should: [
-                        {
-                          simple_query_string: {
-                            query: q
-                          }
-                        },
-                        {
-                          query_string: {
-                            query: "*" + q + "*"
-                          }
-                        }
-                      ],
-                      minimum_should_match: 1
-                    }
+      if (q.length == 26){
+        var start_date = q.substr(5,10);
+        var stop_date = q.substr(16,10);
+        try {
+          const results = await this.client.search({
+            index: this.config.indexName,
+            body: {
+              query: {
+                range: {
+                  createdAt: {
+                    gte: start_date,
+                    lte: stop_date
                   }
-                ]
-              }
-            },
-            from: 0,
-            size: 50,
-            _source: ['title', 'description', 'path', 'locale', 'createdAt'],
-            suggest: {
-              suggestions: {
-                text: q,
-                completion: {
-                  field: 'suggest',
-                  size: 5,
-                  skip_duplicates: true,
-                  fuzzy: true
+                }
+              },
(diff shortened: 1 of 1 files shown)


server/modules/search/elasticsearch/engine.js
```diff
@@ -87,63 +84,156 @@ module.exports = {
...
-                        }
-                      },
-                      {
-                        query_string: {
-                          query: "*" + q + "*"
-                        }
-                      }
-                    ],
-                    minimum_should_match: 1
+    if(q.search("date:")==0){
```
NGPixel, line 87: Use `.indexOf` instead of `.search`. You're not using regex which makes it significantly slower than indexOf. Also never use `==`, always `===` unless you're specifically looking for null.

server/modules/search/elasticsearch/engine.js
```diff
@@ -2,9 +2,7 @@ const _ = require('lodash')
 const stream = require('stream')
 const Promise = require('bluebird')
 const pipeline = Promise.promisify(stream.pipeline)
-
```
NGPixel, line 5: Please restore all the empty lines that are removed in this PR. Makes the code easier to read...

server/modules/search/elasticsearch/engine.js
```diff
@@ -87,63 +84,156 @@ module.exports = {
...
-                          query: "*" + q + "*"
-                        }
-                      }
-                    ],
-                    minimum_should_match: 1
+    if(q.search("date:")==0){
+      // get the date range
+      if (q.length == 26){
+        var start_date = q.substr(5,10);
+        var stop_date = q.substr(16,10);
```
NGPixel, line 91: Avoid using exact character indexes. Check instead of the presence of a comma and split the string if present. This allows for greater flexibility in the format.

Also, the ISO 8601 standard for date range uses a `/` as a separator.

```suggestion
      if (q.indexOf(',') > 0) {
        const dateRange = q.substring(5).split('/')
```

Finally, never use `var`, always `let` or `const`. `var` is globally scoped and unsafe.

server/modules/search/elasticsearch/engine.js
```diff
@@ -87,63 +84,156 @@ module.exports = {
...
-        results: _.get(results, 'body.hits.hits', []).map(r => ({
-          id: r._id,
-          locale: r._source.locale,
-          path: r._source.path,
-          title: r._source.title,
-          description: r._source.description
-        })),
-        suggestions: _.reject(_.get(results, 'suggest.suggestions', []).map(s => _.get(s, 'options[0].text', false)), s => !s),
-        totalHits: _.get(results, 'body.hits.total.value', _.get(results, 'body.hits.total', 0))
       }
```
NGPixel, line 174: You have 2 if with a condition but not else block. What happens if I type `date:abc`, it never returns.

NGPixel (changes requested):
You should allow for a normal query to be added after the date parameter, e.g.:
```
date:2020-05-15/2020-06-14 foo bar
```
If the query startsWith `date:` and contains a space, then the first part should be decomposed for date range while the rest is a standard query string.

Issue #2544, opened 2020-10-05
Azure Active Directory - Error 500 on callback


LtrDan:
[docker version 2.5.159]
Hey,did **logout** from AD Azure - after trying to login again - passing microsoft login and redirected to 
https://ourwikidomain/login/azure/callback

Getting error 500 "Oops,Something went wrong"
Happens on all organization's computers
Going to Incognito - works..

maybe some cache problem? after logging out?

Thanks, Dan

Issue #1992, opened 2020-06-04
Azure AD login fails with "Invalid email / username or password."


Jamesits:
**Describe the bug**

**To Reproduce**
1. Set up Azure AD auth according to [doc](https://docs.requarks.io/en/auth/azure)
2. Enable self-registration
3. Try to log in with Azure AD

**Expected behavior**
I should be able to log in

**Screenshots**

**Host Info (please complete the following information):**
 - OS: Docker
 - Wiki.js version: 2.3.81
 - Database engine: MariaDB 10.4.12-MariaDB-1:10.4.12+maria~buster-log

**Additional context**

JWT claim looks good to me. The `email` and `preferred_username` both exist but are different. 

Web page log:
```
Oops, something went wrong...
Invalid email / username or password.
AuthLoginFailed 1002: Invalid email / username or password.
    at /wiki/server/models/users.js:282:38
    at allFailed (/wiki/node_modules/passport/lib/middleware/authenticate.js:107:18)
    at attempt (/wiki/node_modules/passport/lib/middleware/authenticate.js:180:28)
    at Strategy.strategy.fail (/wiki/node_modules/passport/lib/middleware/authenticate.js:302:9)
    at Strategy.failWithLog (/wiki/node_modules/passport-azure-ad/lib/oidcstrategy.js:1498:15)
    at /wiki/node_modules/passport-azure-ad/lib/oidcstrategy.js:662:21
    at /wiki/node_modules/passport-azure-ad/node_modules/async/lib/async.js:52:16
    at Immediate._onImmediate (/wiki/node_modules/passport-azure-ad/node_modules/async/lib/async.js:1206:34)
    at processImmediate (internal/timers.js:456:21)
    at process.topLevelDomainCallback (domain.js:137:15)
```

Console log:
```
2020-06-04T09:50:38.011Z [MASTER] info: Authentication Strategy azure: [ OK ]
2020-06-04T09:50:38.012Z [MASTER] info: Authentication Strategy github: [ OK ]
2020-06-04T09:50:38.012Z [MASTER] info: Authentication Strategy local: [ OK ]
```

Smankusors:
What do you mean by :
> The `email` and `preferred_username` both exist but are different.

?

Jamesits:
@Smankusors There are both `email` and `preferred_username` claims but their values are different.

smg-bg:
Same here...

[...]

PR #2567, opened 2020-10-11
Enable passport-azure-ad workaround for SameSite cookies

yaegashi:
This adds cookieEncryptionKeyString configuration in the Azure AD authentication module.  It represents an array of cookie encryption strings and enables workaround for SameSite cookies.

See [the passport-azure-ad doc](https://github.com/AzureAD/passport-azure-ad#41-oidcstrategy) about new `OIDCStrategy` options: `cookieSameSite`, `useCookieInsteadOfSession`, `cookieEncryptionKeys`.  It's not mentioned in the doc that `cookieSameSite` won't work without `useCookieInsteadOfSession` or without `cookieEncryptionKeys`.

When enabled, it returns a short-lived cookie with `Secure; SameSite=None` during Azure AD OpenID Connect authentication attempts.

image

Unfortunately there's no unit test because I'm not familiar with auth module testing.  And it's missing the way to add this to existing Azure AD authentication configurations (direct fixes on DB?).

This may fix the problem #2544, #1992

### DIFF
--- a/server/modules/authentication/azure/authentication.js
+++ b/server/modules/authentication/azure/authentication.js
@@ -10,6 +10,19 @@ const OIDCStrategy = require('passport-azure-ad').OIDCStrategy
 
 module.exports = {
   init (passport, conf) {
+    // Workaround for Chrome's SameSite cookies
+    // cookieSameSite needs useCookieInsteadOfSession to work correctly.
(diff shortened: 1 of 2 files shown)


NGPixel (approved):

michalcesek:
I tried to apply this workaround but it didn't help. I still have to clear the cookies.

PR #2579, opened 2020-10-14
Reduce chunk size to circumvent sqlite restriction

### LINKED ISSUE #1503: SQLITE_ERROR: too many SQL variables
**Describe the bug**
When rebuilding page tree, with SQL Lite I get an error.

**To Reproduce**
1. Import from Wikijs 1.
2. Start up and you'll get error
3. Go to utilities and then to Rebuild Page Tree (and you'll get the same error)

**Expected behavior**
No error

**Screenshots**
Current I'm getting an error that says 

```
insert into pageTree (depth, id, isFolder isPrivate, localeCode, pageId, parent, path, privateNS, title)
select 1 as depth, false as isfolder, 0 as isPrivate, en as localCode, 1 as pageID, null as parent, .....
Union all select ...
union all select ...
```
And so on forever, followed by,

    SQLITE_ERROR: too many SQL variables

This is not a good a way to create a select statement. I believe that you should follow the form of 

    INSERT INTO table ( col1, col2 ) VALUES (1,2), (2,3);

The SQL statement generated I believe is too massive as-is.

pchinery:
This fixes #1503

### DIFF
--- a/server/jobs/rebuild-tree.js
+++ b/server/jobs/rebuild-tree.js
@@ -57,7 +57,7 @@ module.exports = async (pageId) => {
     await WIKI.models.knex.table('pageTree').truncate()
     if (tree.length > 0) {
       // -> Save in chunks, because of per query max parameters (35k Postgres, 2k MSSQL, 1k for SQLite)
-      for (const chunk of _.chunk(tree, 100)) {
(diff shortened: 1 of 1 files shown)


NGPixel:
This was fixed by another PR so closing this one.

PR #2588, opened 2020-10-16
Add quotation mark to metadata injection to avoid YAML parse exception.

loudcolour:
Quoting & unquoting page.title and page.description for YAML metadata is required for avoiding parse exception when page.title or page.description contains special letters (emojis, CJK letters, etc.). Otherwise, it will cause an error when importing the .md file to Wiki.js' database.

### DIFF
--- a/server/helpers/page.js
+++ b/server/helpers/page.js
@@ -73,8 +73,8 @@ module.exports = {
    */
   injectPageMetadata(page) {
     let meta = [
-      ['title', page.title],
-      ['description', page.description],
+      ['title', '"' + page.title + '"'],
+      ['description', '"' + page.description + '"'],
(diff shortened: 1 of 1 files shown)


server/helpers/page.js
```diff
@@ -73,8 +73,8 @@ module.exports = {
    */
   injectPageMetadata(page) {
     let meta = [
-      ['title', page.title],
-      ['description', page.description],
+      ['title', '"' + page.title + '"'],
+      ['description', '"' + page.description + '"'],
```
NGPixel, line 77: Would be a good idea to escape both variables in case they contain " in them as well.

NGPixel (changes requested):

PR #2628, opened 2020-10-27
fix: use absolute URL for logo in email

EricFromCanada:
Prepends the wiki site URL to the logo path used when sending email from the wiki.

Before: `<img src="/uploads/logo.png"`
After: `<img src="http://wiki.site.url/uploads/logo.png`

### DIFF
--- a/server/core/mail.js
+++ b/server/core/mail.js
@@ -56,7 +56,7 @@ module.exports = {
       subject: `${opts.subject} - ${WIKI.config.title}`,
       text: opts.text,
       html: _.get(this.templates, opts.template)({
-        logo: WIKI.config.logoUrl,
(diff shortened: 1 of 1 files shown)


server/core/mail.js
```diff
@@ -56,7 +56,7 @@ module.exports = {
       subject: `${opts.subject} - ${WIKI.config.title}`,
       text: opts.text,
       html: _.get(this.templates, opts.template)({
-        logo: WIKI.config.logoUrl,
+        logo: WIKI.config.host + WIKI.config.logoUrl,
```
NGPixel, line 59: logoUrl can also be an absolute URL. There should be a check to see whether logoUrl starts with http and only prefix the host if not.

NGPixel (changes requested):

EricFromCanada:
Added a ternary that conditionally prepends the host URL. Tested both cases on my installation.

NGPixel (approved):

PR #2645, opened 2020-10-31
fix: inline math interpreted as attributes

### LINKED ISSUE #1581: Katex subscript rendering issue
## Description
Katex rendering seems to have a bug with rendering subscript.

## To reproduce:
Bug occurrs when trying to use more than one character in subscript.
I used the Katex code `P_{static}=I_{DD}V_{DD}` and while it was rendered partly correct.
Another example would be `V_{DD}=5V`
The logs output warnings like this:
`2020-03-12T16:45:06.500Z [JOB] warn: KaTeX parse error: Expected group after '_' at position 2: I_`
`2020-03-12T16:45:06.501Z [JOB] warn: KaTeX parse error: Expected group after '_' at position 19: …static}=I_{DD}V_`

## Expected Behaviour
Katex code should have same output as on [katex.org](https://katex.org/)

## Host Info:
 - OS: Docker
 - Wiki.js version: 2.1.113
 - Database engine: postgres 11.7

edomora97:
When using inline math (`$e^{-x^2}$`) the curly braces are interpreted as attributes by [markdown-it-attrs](https://github.com/arve0/markdown-it-attrs). Since most of the times they are not valid attributes they simply get removed.

This patch escapes the curly braces (the default attribute delimiter), fixing the KaTeX rendering errors.

The problem occurs only when the inline math ends with a _curly brace block_ (see examples below). For this reason the third example below works also in the current version.
An alternative fix may have been to force the math block not to end with the curly braces, for example by adding some spacing (e.g. `\,`). I preferred not adding some spurious spaces, the escaping strategy should be pretty invisible.

One could think it should be enough to escape the last `{` and `}`, in practice it's not enough: `$e^{e^{x}+1}$` would become `$e^{e^{{x}+1}}$`, which is valid but clearly not equivalent. (note that this example is not broken in the current version since the last _curly block_ contains another _curly block_, which apparently it's fine).

It would be nice to simply skip that rule for `katex_inline` block types but as far as I know markdown-it-attrs doesn't have such an [option](https://github.com/arve0/markdown-it-attrs/blob/3254c2bb5f091a7680e8ac86ce62dd4561fb8e89/patterns.js#L306).

---

### Example
```
$e^{-x^2}$

$P_{static} = I_{DD}V_{DD}$

$V_{DD}=5V$
```

### Previously
image

```
ParseError: KaTeX parse error: Expected group after '^' at position 2: e^̲
ParseError: KaTeX parse error: Expected group after '_' at position 21: …atic} = I_{DD}V_̲
```

### Now
image

------

Fixes #1581

### DIFF
--- a/client/components/editor/common/katex.js
+++ b/client/components/editor/common/katex.js
@@ -81,7 +81,14 @@ export default {
     if (!silent) {
       token = state.push('katex_inline', 'math', 0)
       token.markup = '$'
-      token.content = state.src.slice(start, match)
+      token.content = state.src
+        // Extract the math part without the $
+        .slice(start, match)
+        // Escape the curly braces since they will be interpreted as
+        // attributes by markdown-it-attrs (the "curly_attributes"
+        // core rule)
+        .replaceAll("{", "{{")
+        .replaceAll("}", "}}")
     }
 
     state.pos = match + 1


NGPixel (approved):

ivlydp:
It seems still have this problem in version 2.9.191
for inline function $H_{ADAS2Eye}$
` KaTeX parse error: Expected group after '_' at position 2: H_̲` 
please help check it.

PR #2751, opened 2020-11-24
Interactive Plantuml with searchable content

bharatrajagopalan:
This is a copy of the original pull request [here](https://github.com/Requarks/wiki/pull/1337) 

Creating this pull request on dev branch as discussed with @NGPixel 

Summary:
Plantuml is currently created as an image which means that links donot work and content is treated as a picture.

This PR does the following:
- Renders plantuml as svg using `object` tag so that it is interactive and selectable (e.g. you can now copy paste text from within the  diagram
- Places the original plantuml content inline inside the object tag so that wikijs search indexes the content for searchability.

As `object` is typically insecure, it will go through the following checks and sanitisation before being kept
- Only attributes that are required for plantuml are allowed. 
- If any of the required attributes are missing then the node is discarded
- the `data` attribute must start with the URL/sub domain of preconfigured plantuml server and image type (e.g. svg) in wikijs config (e.g. https://plantuml.requarks.io/svg) 
-  if the above checks pass then allowed attributes other than `data` are reset to required values for plantuml processing

Tested using [sample markdown](https://github.com/Requarks/wiki/files/4005764/Test-Plantuml.txt)


bharatrajagopalan:
Closing this pull request - will squash the commits to make it more readable and open a new one

PR #2754, opened 2020-11-24
Changes for interactive plantuml

bharatrajagopalan:
This is a copy of the original pull request [here](https://github.com/Requarks/wiki/pull/1337) 

Creating this pull request on dev branch as discussed with @NGPixel . I had originally created this a while ago by forking the  master branch and hence needed to rewrite this using DOMPurify.

Summary:
Plantuml is currently created as an image which means that links do not work and the plantuml content is treated as a picture and is not searchable

This PR does the following:
- Renders plantuml as svg using `object` tag so that it is interactive and selectable (e.g. you can now copy paste text from within the  diagram) with a backup non-interactive img tag if browser will not render `object`
- Places the original plantuml content inline inside the object tag to ensure that this is not visible but included in wikijs search index

As `object` tag is typically insecure, it is checked and sanitised to make sure that 
- Only the configured plantuml server, image format (e.g. https://plantuml.requarks.io/svg/<encoded plantuml>) are allowed in the url parameter passed in `data`; Any other value results in the node being discarded
- Only attributes required for plantuml+wikijs are allowed (`style`/`class`/`type`/`data`/`alt`); mismatch in number of attributes being passed in will result in the node being discarded. 
- Allowed attributes  other than `data` are overriden to the required values for plantuml to work with wikijs explicitly to avoid any code injection scenarios

Tested using [sample markdown](https://github.com/Requarks/wiki/files/4005764/Test-Plantuml.txt)

### DIFF
--- a/server/modules/rendering/html-security/renderer.js
+++ b/server/modules/rendering/html-security/renderer.js
@@ -32,6 +32,59 @@ module.exports = {
         allowedAttrs.push('allow')
       }
 
+      //Changes to keep interactive plantuml object tag
+
+      //only allow specific attributes for plantuml object node
+      allowedTags.push('object')
+      allowedAttrs.push('data')
+      allowedAttrs.push('type')
+      allowedAttrs.push('style')
+      allowedAttrs.push('class')
+      allowedAttrs.push('alt')
+
+      DOMPurify.addHook('uponSanitizeElement',  (node, data) => {
+        // keep object node only if it is
+        // authorised plantuml using the configured plantuml server
+        // force attribute values to configured params
+        // insert the plantuml inside the object as text for search
+        let isPumlNode=false
+
+        if (data.tagName === 'object') {
+          //console.log ("Found object node - validating")
+          //remove node if it doesn't conform to plantuml structure
+          if (!( 'data' in node.attributes
+                && 'class' in node.attributes
+                && 'style' in node.attributes
+                && 'type' in node.attributes
+                && 'alt' in node.attributes)
+          ) {
+            //console.log ("Attribute mismatch - removing object node")
+            return node.parentNode.removeChild(node)
+          }
+
+          dataAttribute = node.getAttribute ('data')
+
+          //only allow configured plantuml server and image format in url
+          if (dataAttribute
+              && dataAttribute.startsWith(`${pumlServer}/${pumlImageFormat}`)
+          ) {
+            //console.log ("Plantuml node found - setting atribute values")
+            isPumlNode=true
+            node.setAttribute ('type', `${pumlObjectType}`)
+            node.setAttribute ('style', `${pumlObjectStyle}`)
+            node.setAttribute ('class', `${pumlObjectClass}`)
+            node.setAttribute ('alt', '')
+          }
+
+          //if not a plantuml node, then sanitise it
+          if (!isPumlNode) {
+            console.log ("Removing unknown object node")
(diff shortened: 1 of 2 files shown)


server/modules/rendering/html-security/renderer.js
```diff
@@ -32,6 +32,59 @@ module.exports = {
         allowedAttrs.push('allow')
       }
 
+      //Changes to keep interactive plantuml object tag
```
NGPixel, line 35: Should be an option, set to Off by default, as this opens up potential security issues.

NGPixel (changes requested):

server/modules/rendering/html-security/renderer.js
bharatrajagopalan: Hi @NGPixel 

Currently, the allowedAttrs/allowedTags is being used to trigger logic in the uponSanitizeElement logic to remove the node completely if it doesn't meet the criteria for plantuml.

This logic validates that the url is set to the preconfigured plantuml server and overwrites atrribute values being passed to acceptable values, else rejects the node altogether

I tried to see if there was a hook that let me parse the content and adjust it without setting the `allowAttributes` `allowTag` (e.g. `beforeSanitizeElement` however in that scenario, DOMpurify would always just remove the node.

So I had a read of use cases here: https://github.com/cure53/DOMPurify/blob/main/demos/README.md#hook-to-allow-and-sand-box-all-javascript-link

It does seem to suggest that we first allow the attributes and then use logic in uponSanitizeElement to remove them if they don't meet the criteria - this is what I have done.

Note that we do need to use an object node to make sure that the svg interactive i.e. to be select and copy text from the diagram or to be able to click links, the above steps ensure that the tag is secure by only allowing access to the preconfigured plantuml server and forcing other attribute values.

Detailed logic is
- Examine  all the plantuml tags and attributes;  To this I need to use `allowedTags` and `allowedAttrs` 
- In `uponSanitizeElement` - the node is rejected if it doesn't satisfy the criteria of a valid plantuml node i.e.
 - if the number of attributes doesn't match
 - if the url doesn't point to the configured plantuml server in wikis settings
- once the node is determined to be a plantuml node, the allowed attributes (as mentioned above) are forcefully overwritten to predefined values to make sure that no code injection is possible
- If no match is found, then remove the node

bharatrajagopalan: @NGPixel  see changes in  commit [571d1ddf0e2ead23ee1cb7691555fa7e1a81bd13](https://github.com/Requarks/wiki/pull/2754/commits/571d1ddf0e2ead23ee1cb7691555fa7e1a81bd13) 

Summary:
- Object tag and attributes are allowed only if plantuml `svg` mode is selected. User already has the option to select png as an alternative so this is optional.
- To ensure that the user doesn't lose out too much by selecting png as the format - the plantuml content is now stored in hidden text in the html so that it is still searchable irrespective of the mode that the user selects (png or svg). However obviously user loses the interactiveness that he gets from selecting svg as the format (e.g. links/copy paste text).

bharatrajagopalan: @NGPixel  just bumping this as a follow up - I believe I have incorporated your review comments, so please shout if you need any additional changes

DraakBZH:
any idea when it could be merged? I need this! :D

hernrup-modelon:
+1 for this. The interactivity is a must for plantuml-C4.

PR #2793, opened 2020-12-07
Feature: Custom configuration for elasticsearch

scienceasdf:
For better search results especially in Chinese, which the standard token analyzer may not work well. In Chinese, the adminstrator may need to set the token analyzer to smartcn, itk, etc. This PR adds the elastic analyzer configuration to the search module.

### DIFF
--- a/server/modules/search/elasticsearch/engine.js
+++ b/server/modules/search/elasticsearch/engine.js
@@ -57,12 +57,12 @@ module.exports = {
           const idxBody = {
             properties: {
               suggest: { type: 'completion' },
-              title: { type: 'text', boost: 10.0 },
-              description: { type: 'text', boost: 3.0 },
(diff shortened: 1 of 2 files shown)


server/modules/search/elasticsearch/engine.js
```diff
@@ -63,14 +63,23 @@ module.exports = {
               locale: { type: 'keyword' },
               path: { type: 'text' },
               tags: { type: 'text', boost: 8.0 }
-            }
+            },
```
NGPixel, line 66: Remove dangling comma

NGPixel (changes requested):

server/modules/search/elasticsearch/engine.js
scienceasdf: Fix that

NGPixel (approved):

PR #2933, opened 2021-01-20
Change "Site Tree" navigation to redirect to folder landing pages

douglaslassance:
After deploying Wiki.js in our company, we had a lot of feedback from confused users when using the "Site Tree" navigation. This revealed to be due to a couple of interaction facts:

* Clicking on a folder would not take you to the landing page of that folder, requiring a second click on the landing page to reach the information.
* Clicking on the landing page would take the navigation back to the level above the landing page folder, making it impossible to read the landing page while seeing the sub-pages of that folder in the nav.

We have explored the other types of navigation but the ability for the "Site Tree" to dynamically display sub-pages was just too valuable. As a result, we decided to patch it by making sure that **clicking on a folder would redirect you to that folder's landing page**. We also modified the tree to represent the home page as the root and added a redirect to it.

We have no hope or desire to see this request merged in its current state since we are aware of the patched nature of the change. That being said, we thought this could be a great platform to discuss the validity of the feature and the way you'd see it implemented. We obviously would like to avoid having to maintain a fork and are ready to put in the necessary work.

Thank you.

### DIFF
--- a/client/themes/default/components/nav-sidebar.vue
+++ b/client/themes/default/components/nav-sidebar.vue
@@ -46,18 +46,14 @@
     //-> Browse
     v-list.py-2(v-else-if='currentMode === `browse`', dense, :class='color', :dark='dark')
       template(v-if='currentParent.id > 0')
-        v-list-item(v-for='(item, idx) of parents', :key='`parent-` + item.id', @click='fetchBrowseItems(item)', style='min-height: 30px;')
+        v-list-item(v-for='(item, idx) of parents', :key='`parent-` + item.id', @click='routeNavLink(item)', style='min-height: 30px;', :input-value='path === item.path')
           v-list-item-avatar(size='18', :style='`padding-left: ` + (idx * 8) + `px; width: auto; margin: 0 5px 0 0;`')
             v-icon(small) mdi-folder-open
           v-list-item-title {{ item.title }}
         v-divider.mt-2
-        v-list-item.mt-2(v-if='currentParent.pageId > 0', :href='`/` + currentParent.path', :key='`directorypage-` + currentParent.id', :input-value='path === currentParent.path')
-          v-list-item-avatar(size='24')
-            v-icon mdi-text-box
-          v-list-item-title {{ currentParent.title }}
-        v-subheader.pl-4 {{$t('common:sidebar.currentDirectory')}}
+        v-subheader.pl-4(v-if='currentItems.length') {{$t('common:sidebar.currentDirectory')}}
       template(v-for='item of currentItems')
-        v-list-item(v-if='item.isFolder', :key='`childfolder-` + item.id', @click='fetchBrowseItems(item)')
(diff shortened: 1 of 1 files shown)


thoni56:
This would be very valuable as a standard or alternative to the current tree navigation which to many is non-intuitive.

It might be an evidence of the fact that there (in my humble opinion) probably is a deeper, almost philosophical, discussion to be had to make Wiki.js the best wiki, but that should probably be taken somewhere else. On the feature request list?

thoni56:
I've tried this PR out and I must say it pretty much nails it. I'm not sure why you consider it of "patched nature" ;-)

It does change some earlier behaviour, yes, so I guess it needs to be discussed in view of that change, but to us it fits the bill and would be happy to see this in.

douglaslassance:
I think the main concern is that this should be an option. Either an entirely new navigation mode, either an option on the Site Tree mode. Since that's a bit of additional work, we wanted to make sure that @NGPixel is board conceptually.

thoni56:
Personally, I can't see how anyone could not see this as an improvement on the current tree navigation, but again, that's just me ;-)

I totally agree that we need @NGPixel onboard here, especially since I've seen a few other aspects that probably would benefit from a change in this direction (where new pages are assumed to end up, what page a link goes to by default when you click on a folder, ...) But all those can be completely separate, smaller, changes.

Me and a few others at my company might be willing to do what we can to help if this is the direction Wiki.js navigation is going.

NGPixel:
This will be the default behavior in 3.0.

gmasil:
@douglaslassance Thank you very much for this PR. I merged it into my fork and it will definitely stay!
Great improvement for me as I am still trying to convince myself to move from Confluence to Wikijs, now that Atlassian is not interested in smaller dev teams anymore.
Once again, thank you very much!

Douglasproglima:
@NGPixel What will be the release date of version 3.0?

NGPixel:
> @NGPixel What will be the release date of version 3.0?

See https://docs.requarks.io/releases

digitalLumberjack:
One other instance with this feature :D

Thanks @douglaslassance !

Thanks a lot for wikijs @NGPixel and courage for the 3.0 ! :rocket:

jaimehrubiks:
This is the only thing missing in WikiJS

jonaskello:
Also found this really confusing. I was thinking it was some kind of bug or that I did something wrong in the setup, but now I found this PR I understand it is confusing by design :-). Is there some easy way to introduce this patch into a container (k8s) setup?

gmasil:
@jonaskello Just create a fork, [merge this PR into your fork](https://stackoverflow.com/questions/6022302/how-to-apply-unmerged-upstream-pull-requests-from-other-forks-into-my-fork) and then build the docker image on your own as described in the [wiki](https://docs.requarks.io/dev). If you don't need a fork or version control just checkout the repository and apply the patch of this PR from here: [2933.patch](https://patch-diff.githubusercontent.com/raw/requarks/wiki/pull/2933.patch) using the git command `git apply 2933.patch`. Then build the docker image.

Issue #2572, opened 2020-10-12
Rocket Chat Authentication Instructions?


HiveMindNet:
**Question**
Trying to use the Rocket Chat auth but no documentation anywhere to help me test i. 

**Host Info (please complete the following information):**
OS: Latest Docker
Wiki.js version: 2.5.159
Database engine: Latest Docker PG_VERSION	11.8

Can someone please either point bee to some docs or help me understand where I get or set up the 'Client ID' and 'Client Secret' in my Rocket admin? Also, will this integration still work if I'm actually using iFrame auth on my Rocket instance? (Rocket authenticates from our other main website).

Any pointers or docs would be greatly appreciated :)

Thank you in advance.

HiveMindNet:
Has anyone got this working at all?

rychergates:
Have you looked at these docs https://docs.rocket.chat/guides/administrator-guides/authentication

HiveMindNet:
Hi - Yes, that all refers to authenticating Rocket from another site. Not authenticating other sites from Rocket.

HiveMindNet:
Has *Anyone* managed to get Rocket Chat Authentication working? If so could you please share how you set it up?

erikvanberkum:
i got it installed to the point that i get the following error in wiki.js

Oops, something went wrong...
Missing or invalid email address from profile

![image](https://user-images.githubusercontent.com/7335299/107193974-cfac0b80-6a32-11eb-8a77-9948fa0973d8.png)

deg0nz:
I have the exact same issue. 

I wanted to debug this further, but I can't find any method to increase the log level. I found no docs about some log level environment variable and mounting `config.yml` with increased `loglevel` seems not to work as mentioned in #2497.

What I found was, that the request to Wiki.js's auth callback url returns a HTTP 500.

deg0nz:
Okay, I digged into this a little and found the following:

[...]

PR #3122, opened 2021-02-22
Fix Rocket.Chat auth module

deg0nz:
As mentioned in #2572, Wiki.js uses the wrong property to access a users' E-Mail address in the Rocket.Chat auth module.

This PR fixes this.

I have tested this change with the current `dev` branch on one of my servers and it's working.

Rocket.Chat's response contains an array containing all the emails a user has added to Rocket.Chat. The currently used property `email` does not exist in the response. Please see https://docs.rocket.chat/api/rest-api/methods/authentication/me.

**Since we have no option to choose  which E-Mail to use for the OAuth mechanism, I just assumed, we take the first one in the array and go for it.**

Please tell me, if you are OK with this, since it can be considered as a dirty hack. 
I can add some validation that checks, if the array is empty that throws an error or something.

### DIFF
--- a/server/modules/authentication/rocketchat/authentication.js
+++ b/server/modules/authentication/rocketchat/authentication.js
@@ -23,7 +23,7 @@ module.exports = {
           cb(null, {
             id: usr._id,
             displayName: _.isEmpty(usr.name) ? usr.username : usr.name,
-            email: usr.email,
+            email: usr.emails[0].address,
             picture: usr.avatarUrl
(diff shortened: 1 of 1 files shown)


deg0nz:
@NGPixel is there a specific reason for this to be held back?

I would like to update my wiki.js server, but the update depends on this fix because my users are using rocketchat auth.

NGPixel (approved):

Issue #2797, opened 2020-12-08
Network error: Response not successful: Received status code 413


beerygaz:
**Question**
I get the above error when trying to save a large page.  The page is large due to an innserted draw.io image, but removing the tags and just pasting the large block of (what appears to be) base64 text into the page generates the same error.  The total size of the page is 1.2Mb

I have set the Admin/Security/MaxUploadSize to 5242880 bytes (5Mb) but still see this error.

I have eliminated my reverse proxy (Trefik) by bypassing the proxy to no avail.

I can confirm that if I reduce the page size from 1.2Mb to 770k it will save, so the issues exists somewhere in that range, but I don't know where that constraint exists.

**Host Info (please complete the following information):**
OS: Docker
Wiki.js version: 2.5.170
Database engine: git

NGPixel:
the max upload size is for assets only, not for page content, so changing it won't have any effect.

The body parser for requests is limited to 1Mb. Draw.io diagrams can quickly cause the page to go over that limit. It will be addressed in a future release, but in the meantime you can increase the limit manually by changing this line:

https://github.com/Requarks/wiki/blob/a1f1cd7356ddc2ba9054aa0f39b0b9f3f43ce3f8/server/master.js#L94

bestwnh:
Hi @NGPixel,

I like to contribute this issue. It prevents me to create some pages with base64 images. And my company use it with AWS so it is a bit hard to modify the code. So I just hope I can make sure this can be config in the next version.

But seems like I need to request access to the translation web service first for the locale files before local development right? I tried gitter but all I see is a 404 page. So could you tell me what do I need to do? My gitter account link from my GitHub so just feel free to contact me.

JohnnyDevCraft:
Is there an easy way to handle this in a running container, or will this require a redeploy of a new image?

JohnnyDevCraft:
So I updated the master.js to the following:

`app.use(bodyParser.json({ limit: '10mb' }))`

and I'm still getting a 413 when I try to attach a 1.2 MB PDF to an article.  Any thoughts on what I should try next?

davidib2001:
Same here, tried 
app.use(bodyParser.json({ limit: '50mb' }))
with same result.
Anybody?  Th elack of being able to use visual ais in wiki really limits it's usefulness

TheLouD1:
not sure why @NGPixel closed this as completed on Jan 14, 2023 as it's still very much an issue ??

[...]

PR #3150, opened 2021-02-26
add config for body parser size

bestwnh:
PR for https://github.com/Requarks/wiki/issues/2797
This need to restart node to make it work. The hint should include this. But I don't know how to combine with the locale files. So I just leave it with key and you can add the locale files to control the display text.

I don't familiar with the structure of this project. I just learn by some exist codes. But I do test the result and it work fine for me.

### DIFF
--- a/server/app/data.yml
+++ b/server/app/data.yml
@@ -80,6 +80,7 @@ defaults:
     uploads:
       maxFileSize: 5242880
       maxFiles: 10
+    bodyParserSize: 1
     flags:
       ldapdebug: false
       sqllog: false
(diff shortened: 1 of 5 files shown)


AlexisTonneau:
Great PR ! I forgot to change the hardcoded value after an update and wasn't able to change my pages with schemas...

server/app/data.yml
```diff
@@ -80,6 +80,7 @@ defaults:
     uploads:
       maxFileSize: 5242880
       maxFiles: 10
+    bodyParserSize: 1
```
Roriz, line 83: What do you think to increase this value because of bigger diagrams? Something like this:
```suggestion
    bodyParserSize: 20
```

based on pr: https://github.com/Requarks/wiki/pull/3850#pullrequestreview-848083347

testn:
Can we get this in?

PR #3165, opened 2021-03-03
Error modal when uploading assets with already existing name

### LINKED ISSUE #3159: Already existing asset overridden with no warning when uploading a new asset with the same name
Hi, first thank you and congratulations for this amazing tool! 🎉 

My team noticed an issue with asset uploading, I'd be happy to help fix the code and do a PR if you give me some guidelines.

**Describe the bug**
When uploading a new asset with the same name as another asset, the old asset is overridden without any warning being raised. This issue concerns two assets in the same folder, `/root` for example.

**To Reproduce**
1. In editor mode, click on *Insert Assets*.
2. Upload in `/root` an asset having the name of an already existing asset which is also in `/root`.
3. The old asset is now overridden by the new one.

**Expected behavior**
A warning should be raised when trying to upload an asset with an already existing name in the folder.

PierreGuilmin:
This fixes #3159.

The added snippet compares the names of the assets to upload with the names of the already existing assets in the current folder. If there is a conflict an error modal explains the issue and no file is uploaded (even the one with no names conflict).

Note: If another asset with the same name exists but in a different folder no error modal is raised and the files are uploaded.

This PR need the following translation to be added (currently in `server/locales/en.yml`):
```yml
editor:
  assets.conflictingFileNames: 'Conflicting file names, please rename the uploaded assets.'
```

Demo:

### DIFF
--- a/client/components/editor/editor-modal-media.vue
+++ b/client/components/editor/editor-modal-media.vue
@@ -379,6 +379,14 @@ export default {
           icon: 'warning'
         })
       }
+      const conflictingFiles = _.intersectionWith(files, this.assets, (f, a) => f.filename === a.filename)
+      if (conflictingFiles.length > 0) {
+        return this.$store.commit('showNotification', {
+            message: this.$t('editor:assets.conflictingFileNames'),
+            style: 'error',
+            icon: 'error'
+          })
+      }
       for (let file of files) {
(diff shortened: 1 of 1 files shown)


PierreGuilmin:
Replaced by #3176.

PR #3771, opened 2021-03-26
open folder to show current page when current page is a folder

opalmay:
This was very annoying for me so I made a quick fix.
It works but idk if this is the right way to do this.


digitalLumberjack:
That feature would be really cool !

PR #3821, opened 2021-04-12
fix: graceful shutdown

lkho:
There is a graceful shutdown procedure written in the kernel, but not registered with the process exit handler.

This change adds:
- a SIGINT handler,
- a PM2 specific message handler on windows platform,
- and properly shutdown the scheduler and wait for running jobs to complete before exiting.


NGPixel:
Thanks for all 3 great PRs!

PR #3850, opened 2021-04-16
Increase bodyParser limit

arusa:
Because it is impossible to save a wiki page with a big draw.io diagram it is necessary to increase the bodyParser limit.

See: https://github.com/Requarks/wiki/issues/2797

### DIFF
--- a/server/master.js
+++ b/server/master.js
@@ -91,7 +91,7 @@ module.exports = async () => {
   // GraphQL Server
   // ----------------------------------------
 
-  app.use(bodyParser.json({ limit: '1mb' }))
+  app.use(bodyParser.json({ limit: '20mb' }))
   await WIKI.servers.startGraphQL()
 
   // ----------------------------------------


Extarys:
I would prefer a customizable setting for this.

I found this PR that seems to implement this: https://github.com/Requarks/wiki/pull/3150

I believe this PR can be safely closed.

Roriz (approved):

deggle:
Hi, I'm really confused about this - it seems the limit is very low and there's been threads/talk of this update I've seen going back 2 years but is it not in yet? I'd initially assumed it was linked with the 'max upload' setting but it's not.

I'm unable to get very far with Wiki.js as I need to include some diagrams that have annotations on images. It seems weird to include the draw.io tool, which is a great addition, but be essentially unable to use it for anything more than a very basic diagram.

I tried the workaround (changed the line in the master.js) however this didn't seem to work and then restarting the container rolled back the changes anyway as I guess it won't persist, because docker.

Am I missing something really obvious here?
Or do I need to wait for this PR to go in (and if so, does anyone have an idea when that will be)?

PR #4024, opened 2021-05-27
fix sftp error

### LINKED ISSUE #2297: SFTP Storage not work
**Describe the bug**
Due to the clain

> Base directory where files will be transferred to. The path must already exists and be writable by the user.

I found that not only base dir need to be create manually but also the sub folders do need!
if there is some dir not exists then will come to 

> info: (STORAGE/SFTP) Adding page aaa/bbb/ccc.md...
> warn: No such file

**To Reproduce**
1. Go to setting and set up SFTP 
2. Click on RUN
3. Scroll down to the docker log

**Expected behavior**
subfolder auto create

**Screenshots**

 - OS: Docker
 - Wiki.js version: 2.4.107
 - Database engine: postgres 11.8

**Additional context**

RainshawGao:
when the dir exists, sftp.mkdir() would raise an error and ends the for loop.
fix #2297

### DIFF
--- a/server/modules/storage/sftp/storage.js
+++ b/server/modules/storage/sftp/storage.js
@@ -155,7 +155,12 @@ module.exports = {
         const folderPaths = _.dropRight(filePath.split('/'))
         for (let i = 1; i <= folderPaths.length; i++) {
           const folderSection = _.take(folderPaths, i).join('/')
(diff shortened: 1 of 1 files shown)


RainshawGao:
@NGPixel Any thoughts on this pull request? It would be really nice to get this upstream.

NGPixel (approved):

PR #4156, opened 2021-07-01
fix: default theme summary and detail

n1xx1:
Fix incorrect styling with detail/summary elements when nested together.


broxen:
This follows the standard and is safe to merge.

Further I don't see it having an impact on any existing implementation as specificity hasn't truly changed here

PR #4161, opened 2021-07-04
Add possibility to set PostgreSQL Schema other than public

fredericalix:
I wanted the possibility to use another pg schema than public.

### DIFF
--- a/config.sample.yml
+++ b/config.sample.yml
@@ -29,7 +29,6 @@ db:
   user: wikijs
   pass: wikijsrocks
   db: wiki
-  schema: public # Only for choice a PostgreSQL Schema other than public
   ssl: false
 
   # Optional - PostgreSQL / MySQL / MariaDB only:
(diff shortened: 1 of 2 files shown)


config.sample.yml
```diff
@@ -29,6 +29,7 @@ db:
   user: wikijs
   pass: wikijsrocks
   db: wiki
+  schema: public # Only for choice a PostgreSQL Schema other than public
```
NGPixel, line 32: Put lower down (before sqlite storage option), with comment:
```yaml
# Optional - PostgreSQL only:
schema: public
```

server/core/db.js
```diff
@@ -138,6 +142,10 @@ module.exports = {
           switch (WIKI.config.db.type) {
             case 'postgres':
               await conn.query(`set application_name = 'Wiki.js'`)
+                  // -> Set schema if it's not public             
+                  if (database_schema != 'public') {
```
NGPixel, line 146: Use instead:

```
if (WIKI.config.db.schema && WIKI.config.db.schema !== 'public') {
```

and remove the var declaration above.

NGPixel (changes requested):

NGPixel (approved):

jsprds:
Hi @fredericalix, can I set the pg schema with an ENV variable? Thanks! 🙏

PR #4227, opened 2021-07-22
WIP: Auto installer

wzrdtales:
# Description 

This is an auto installer for installations in a fully automated kubernetes install. A customer needed this b/c he wants to roll it out to all the subsidiaries with the least amount of frictions possible. Since this will be useful for everyone I decided not to do an out of tree implementation.

# Notes

`got` was added as a dependency. This project is still using request, however request is dead and deprecated for long time already. `got` is the `node.js` sides new project that fits into the spot of request best. Especially due to fine grained control over network parameters (such as socket timeouts, ssl timeouts, ...)


wzrdtales:
ping @NGPixel here is the one for dev.

wzrdtales:
@NGPixel ping ping ping

NGPixel:
an auto-install will not be added in 2.x, as this could introduce new issues and would require some extensive documentation as to what it does and what must be provided.

The setup page step has been removed entirely in the upcoming 3.0, so that will no longer we necessary.

PR #4228, opened 2021-07-22
Auto installer (latest branch)

wzrdtales:
Same as #4227 but for the latest branch

# Description 

This is an auto installer for installations in a fully automated kubernetes install. A customer needed this b/c he wants to roll it out to all the subsidiaries with the least amount of frictions possible. Since this will be useful for everyone I decided not to do an out of tree implementation.

# Notes

- `got` was added as a dependency. This project is still using request, however request is dead and deprecated for long time already. `got` is the `node.js` sides new project that fits into the spot of request best. Especially due to fine grained control over network parameters (such as socket timeouts, ssl timeouts, ...) 
- the `setup.js` file was rewritten by my auto fixer, if necessary will adjust that somewhat later, I would suggest you to add a .eslintrc or .prettierrc or both to the project to make sure that your repo standards are well defined and no one has to learn your code style standards and by that decrease interaction.
- the graphql-upload dependency was implicit, but the lock file does not get used during the build. so it was broken and added specifically

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "singleQuote": true
+}
+
--- a/dev/helm/templates/deployment.yaml
+++ b/dev/helm/templates/deployment.yaml
@@ -26,8 +26,12 @@ spec:
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ default "latest" .Values.image.tag }}"
-          imagePullPolicy: {{ default "IfNotPresent" .Values.image.imagePullPolicy }}
+          imagePullPolicy: {{ default "IfNotPresent" .Values.image.pullPolicy }}
           env:
+            {{- range $key, $value := $.Values.env }}
+            - name: {{ $key }}
+              value: {{ $value | quote }}
+            {{- end }}
             - name: DB_TYPE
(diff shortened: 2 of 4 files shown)


NGPixel:
Please remove all the code styling changes that were made that are not specific to this PR (spaces before function params, arguments on multiple lines, etc.).

wzrdtales:
can you please put a code style definition in your project? I really hate to waste time on CS and I don't do any, auto fixer does and it did according to your configured .eslintrc, so it did comply to the rules.

wzrdtales:
i would on top suggest you to add something like husky to the project to make sure that code style automatically gets fixed all the time.

wzrdtales:
small follow up @NGPixel

wzrdtales:
little follow up again :), guess your loaded right now @NGPixel

wzrdtales:
follow up @NGPixel

NGPixel:
Sorry, just noticed this PR is targeting the latest branch, which is wrong. It should target the dev branch.

So either change this PR to point to dev and close your other PR, or apply the changes to your other PR and close this one.

wzrdtales:
there is the same merge request for dev

wzrdtales:
@NGPixel see https://github.com/Requarks/wiki/pull/4227

Issue #1180, opened 2019-11-05
Admin Search bug


sbonaime:
Before searching, as I have 56 pages, I have 4 pages to see in the result list. When I search something like "aaa", I still have at the bottom of the search the 4 pages icons even if I don't have any results ! 

![Capture d’écran 2019-11-05 à 05-11-19 11 25 16](https://user-images.githubusercontent.com/750709/68199985-5025c600-ffbf-11e9-9df0-7fb08a5d8983.png)

thilak007:
@NGPixel , I would like to work on this bug as my first contribution.

craph:
Hello @NGPixel I think you can close this issue because you already accept this PR https://github.com/Requarks/wiki/pull/4280

Best regards, 😃

PR #4280, opened 2021-08-08
fix: resolve admin pages pagination bug

tsh96:
Solve Admin Search bug #1180

### DIFF
--- a/client/components/admin/admin-pages.vue
+++ b/client/components/admin/admin-pages.vue
@@ -61,6 +61,7 @@
             sort-by='updatedAt',
             sort-desc,
             hide-default-footer
+            @page-count="pageTotal = $event"
           )
             template(slot='item', slot-scope='props')
               tr.is-clickable(:active='props.selected', @click='$router.push(`/pages/` + props.item.id)')
@@ -89,6 +90,7 @@ export default {
       selectedPage: {},
       pagination: 1,
       pages: [],
+      pageTotal: 0,
       headers: [
         { text: 'ID', value: 'id', width: 80, sortable: true },
         { text: 'Title', value: 'title' },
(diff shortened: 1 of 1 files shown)


NGPixel (approved):

craph:
Hi @tsh96 ,
I was looking your PR because I'm trying to increase my skills to contribute to this beautiful project and I'm unable to understand what does the `@page-count="pageTotal = $event"` 
I'm not familliar with this synthax for vuejs I think it's pug in a vue template but I'm not aware of what the `@` does 

Please could you give me some advise ?
Thank you very much.

Best regards,

NGPixel:
`@something` is an event in Vue.js. So when event `page-count` is triggered by the component, the expression in quotes is evaluated. In this case, the variable `pageTotal` is updated with the value of $event.

craph:
@NGPixel Thank you very much for the explanation :) 
So if I'm right `@something` is an event and `:something` is a property of the component where it's placed ?

tsh96:
Hi @craph, 
Yes, 
`@` is shorthand for `v-on:` which is used to handle an event.
`:` is shorthand for `v-bind:` which is used to bind a data.

Another well known shorthand is `#` for `v-slot:`

For more information you can refer to 
https://vuejs.org/v2/guide/syntax.html#Shorthands
https://vuejs.org/v2/api/#v-slot

craph:
> Hi @craph, Yes, `@` is shorthand for `v-on:` which is used to handle an event. `:` is shorthand for `v-bind:` which is used to bind a data.
> 
> Another well known shorthand is `#` for `v-slot:`
> 
> For more information you can refer to https://vuejs.org/v2/guide/syntax.html#Shorthands https://vuejs.org/v2/api/#v-slot

Thank you very much @tsh96 :)

Issue #1525, opened 2020-02-25
LDAP Admin Bind Credentials input type


EugeneFenko:
Admin Bind Credentials `input` must be `type="password"`

1. Go to Administration Area > Authentication > LDAP / Active Directory
2. Look on 'Admin Bind Credentials field'
3. See your password

Expected: password as black dots

TimoKruth:
Verified on current `main` (`6f042e97`, 2026-08-06): the requested masking is **still not implemented**.

`server/modules/authentication/ldap/definition.yml` defines `bindCredentials` as a normal `String`. In `client/components/admin/admin-auth.vue`, string strategy properties fall through to the generic `v-text-field` without a `type` attribute. Unlike the dedicated password fields elsewhere in the admin UI, this therefore renders as a plain text input rather than `type="password"`.

NGPixel:
Addressed in v3.

PR #4305, opened 2021-08-14
sensitive ldap bind credentials

ShapedTime:
admin->authentication text fields could be set as sensitive. LDAP bind credentials set as sensitive field.


ShapedTime:
fixes #1525

mesuttalebi:
Is there any plan to Merge this PR soonly?
It is very important to hide ldap admin password from wiki admin users.

NGPixel:
@mesuttalebi Because this PR doesn't address the problem. It merely hides the password on the client side, you can still retrieve it easily via dev tools. The PR also contains completely unrelated changes to the sidebar.

PR #4307, opened 2021-08-15
git storage: handle relocated files

EricFromCanada:
Currently, the git storage model [does not handle files being renamed](https://github.com/Requarks/wiki/discussions/3707). This PR implements this by noticing when a filename in a diff contains ` => ` and relocating the corresponding page.

### DIFF
--- a/server/modules/storage/git/storage.js
+++ b/server/modules/storage/git/storage.js
@@ -142,7 +142,9 @@ module.exports = {
       if (_.get(diff, 'files', []).length > 0) {
         let filesToProcess = []
         for (const f of diff.files) {
-          const fPath = path.join(this.repoPath, f.file)
+          const fMoved = f.file.split(' => ')
+          const fName = fMoved.length === 2 ? fMoved[1] : fMoved[0]
+          const fPath = path.join(this.repoPath, fName)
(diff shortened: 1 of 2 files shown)


server/modules/storage/git/storage.js
```diff
@@ -207,7 +224,23 @@ module.exports = {
...
+          if (assetToRename) {
+            await WIKI.models.assets.query().patch({
+              filename: item.relPath,
+              hash: fileHash
+            }).findById(assetToRename.id)
+            await assetToRename.deleteAssetCache()
+          } else {
+            WIKI.logger.info(`(STORAGE/GIT) Asset was not found in the DB, nothing to rename: ${item.relPath}`)
+          }
+          continue
```
EricFromCanada, line 242: This should theoretically work for assets, but wasn't accessible in my testing because `diff.files` earlier in the file wasn't being populated when a binary was renamed.

NGPixel (approved):

server/modules/storage/git/storage.js
EricFromCanada: This is a limitation of git-js, which doesn't notice binaries being renamed: https://github.com/steveukx/git-js/issues/885

PR #4326, opened 2021-08-19
[Helm-Chart] Feature: config.yml as ConfigMap & offline-mode

Nold360:
This feature adds the `config.yml`-File as a configMap, so it can be modified easier (e.g. we could get rid of the environment variables). 

I also implemented a new values parameter `offlineMode` which will set the `offline` parameter in `config.yml`.


bdsoha:
Looks like this is related to #6057

PR #4327, opened 2021-08-19
Fix post size limitation

SinxLinonZ:
With express module default setting, articles with large size of content (in my case, such as has many diagrams with image data) will fail to upload with a 413 error.
Try to fix this situation by specifying the max body size.

Cheers


NGPixel:
You're duplicating middlewares already defined lower down.

Also, 1024MB is overkill and would open the door to large payload attacks.

PR #4369, opened 2021-09-01
Slack Auth fix: replace passport-slack implementation

ehossack:
This commit replaces the observably defunct @aoberoi/passport-slack
implementation with that of nmaves's passport-slack-oauth2

I have (manually) tested this on a local installation to verify it does correctly work with a slack + wiki.js installation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "node": ">=10.12"
   },
   "dependencies": {
-    "@aoberoi/passport-slack": "1.0.5",
     "@azure/storage-blob": "12.2.1",
     "@exlinc/keycloak-passport": "1.0.2",
     "@joplin/turndown-plugin-gfm": "1.0.27",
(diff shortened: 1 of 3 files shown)


ehossack:
This addresses #3335

ehossack:
Looking at the releases, it looks like 3.0 is the next targeted release. Given that the 2.X stream is currently broken for slack auth, and assuming we're able to create a fix within this PR, is it viable to consider cutting a 2.X release before the 3.0 beta?
(selfish question of course, but I'm not sure the viability of running a patched version myself - given the giant red "THIS IS A DEV VERSION" warning that happens with `node serve` 😅)

nasiralamreeki:
> release 

The 3.0 repo doesn't even look very active so my guess is a beta isn't coming anytime soon. Would be nice to see more updates to 2.x until there is a beta as 2.x still has bugs.

omenos:
Thanks for doing this @ehossack! Hopefully this gets merged and a new 2.x cut, would be really helpful for a project I'm working on.

You might be able to generate the non-dev banner build by following the steps in the `dev/build/Dockerfile` file.

NGPixel (approved):

PR #4376, opened 2021-09-02
chore: add script to package wiki for a local install

ehossack:
This is an attempt to make it easier for folks to be able to run a close-to-production installation off of the main dev branch.

This commit also adds a .dockerignore file to speed up dockerfile builds

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+dist
+node_modules
+.webpack-cache
+assets
+data
--- a/build-local-release.sh
+++ b/build-local-release.sh
@@ -1,0 +1,34 @@
+#! /bin/bash
+
+set -e
+
+### Performs a local build of the project                 ###
+### Useful for when you require a local/forked full build ###
+
+LOCAL_BUILD_VER=${1:-"local"}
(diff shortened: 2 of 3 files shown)


ehossack:
I'm happy if this is rejected, but wanted to start the conversation nonetheless. Is this even the correct set of steps?
If so, should we modify the steps in `.azure-pipelines` in the `buildLinux` phase to just call this script?

One key difference is that instead of doing this command (but adapted for local):
```
find $(Build.StagingDirectory)/wiki/ -printf "%P\n" | tar -czf wiki-js.tar.gz --no-recursion -C $(Build.StagingDirectory)/wiki/ -T -
```

I opted to directly tar from the parent directory via `tar -czf wiki-js.tar.gz wiki/` -- is there a reason why the original command does a find then pipes it to tar?

ehossack:
Not even a comment? 😛

NGPixel:
It was done automatically by GitHub when changing the branch name from dev to main. All other PRs were automatically modified, so not sure why yours didn't.

Try changing the target branch to `main` and I'll re-open it.

PR #4421, opened 2021-09-13
add baecon and error baecon newrelic parameters for eu region support

goto1134:
EU newrelic baecon and error baecon are different from US locale:
```json
{ 
  "beacon": "bam.eu01.nr-data.net",
  "errorBeacon": "bam.eu01.nr-data.net",
  ...
}
```

This PR inroduces baecons as analytics parameters.


NGPixel:
You have a typo in your property names, should be "beacon", not "baecon". Also set the US values as default to help the user identify what it should look like (using `default` property).

goto1134:
@NGPixel, thank you for the suggestions! Can you check now?

NGPixel (approved):

PR #4437, opened 2021-09-14
upgrade helm chart to take in consideration K8S v1.19+ for ingress

RachadAbiChahine:


RachadAbiChahine:
wiki.js can't be installed on v1.19+ without changing the api of the ingress

NGPixel (approved):

chillum:
just wanted to let you know that this comparison doesn't work correctly: on old versions it still tries to use the newest version of the API.

I believe, the reason is recent helm does not contain `.Capabilities.KubeVersion.GitVersion` anymore (it's beed deprecated for a while) and we need to use `.Capabilities.KubeVersion.Version`

PR #4491, opened 2021-09-26
Fix display of Headers within Details Summary

broxen:
Cleaning up Headers within Details Summary to fix an issue with display.

Modified app.scss with the following changes:
- Remove top-margin and keyline under all headers within a Details **Summary**
- Make all Headers within a Details Summary 100% width to fix TOC Anchor location issue.

Let me know if there's anything else I should consider here.

Screenshot fixed:

### DIFF
--- a/client/themes/default/scss/app.scss
+++ b/client/themes/default/scss/app.scss
@@ -816,10 +816,14 @@
       }
       
       > h1, h2, h3, h4, h5, h6 {
-        margin-top: 0;
-        width: 100%;
+        width: 95%;
+        display: inline-block;
         
-        &::after {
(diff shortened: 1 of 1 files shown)


broxen:
This really should be merged with #2502 to accommodate larger summaries. 

I've changed the CSS to accommodate the possibility of summaries containing headers and phrasing content, per the spec: https://html.spec.whatwg.org/multipage/interactive-elements.html#the-summary-element

With this latest revision, the key-line for headers only hides if the header is the _only_ element child of the summary parent.

NGPixel (approved):

PR #4528, opened 2021-10-01
Remove excess Div wrappers

### LINKED ISSUE #4524: Superfluous wrapper divs in generated pages affects css
I'm seeing superfluous empty wrapper divs in my wiki pages. This creates some issues when trying to stylize the first-child or next element using css

**Source:**
```md
## Extra Divs
Test Paragraph here. [link](#)
- list
- list
```
**Generated:**
```html
<h2 id="extra-divs" class="toc-header"><a href="#extra-divs" class="toc-anchor">¶</a> Extra Divs</h2>
<div></div>
<p>Test Paragraph here. <a href="#">link</a></p>
<div></div>
<ul>
  <li>list</li>
  <li>list</li>
</ul>
<div></div>
```

Seems like this section of code is adding divs around hardlines and blank space. I don't see any other utility for this section of code, can it be safely removed?
https://github.com/Requarks/wiki/blob/a20f70ed8de1e5b2e216f6d87b4f8da97019dd56/server/modules/rendering/html-core/renderer.js#L239-L243

broxen:
As noted in #4524 there are an excess of div wrappers in generated page html.

This PR removes this code section because it seems that standalone text is already wrapped in paragraph elements, making this superfluous and simultaneously correcting the issue of excess div wrappers.

Closes #4524

### DIFF
--- a/server/modules/rendering/html-core/renderer.js
+++ b/server/modules/rendering/html-core/renderer.js
@@ -233,16 +233,6 @@ module.exports = {
     })
 
     // --------------------------------
-    // Wrap root text nodes
-    // --------------------------------
-
-    $('body').contents().toArray().forEach(item => {
-      if (item && item.type === 'text' && item.parent.name === 'body') {
-        $(item).wrap('<div></div>')
-      }
-    })
-
-    // --------------------------------
     // Escape mustache expresions
     // --------------------------------
 


NGPixel:
These excess divs are actually there to protect against XSS at the root level in non-div elements. There's probably a more elegant solution but this was a quick way to ensure content is sanitized properly further down the rendering pipeline.

broxen:
>  to protect against XSS at the root level

Ah, I see. I've pushed a proposed change to this so that it ignores "empty" newlines and returns, but still captures potentially dangerous text. I don't know if it's any more elegant, but it works 😄

Tested with the following text in a new page:
```html
<h3>XSS Attack</h3>
<p>Some text here</p> {{constructor.constructor('alert("Wiki.JS Stored XSS")')()&#x7d&#x7d

a
b\n
[][}{}{}<<<<<>>>>>
```

Issue #1304, opened 2019-12-12
asset filename with special characters conflict to markdown syntax


fanlix:
- a file with name such as `a[b).png`
- uploading success,
- then insert into page by markdown syntax : ` ![a[b).png](/a[b).png)`

haha. 
I can change the filename on assets page to fix this.
Should there be a warning before uploading?

TimoKruth:
Verified on current `main` (`6f042e97`, 2026-08-06): the validation/warning requested here is **still missing**.

The upload controller lowercases filenames, replaces whitespace/comma/semicolon/hash characters, and runs `sanitize-filename`. Characters such as `[` and `)` are not rejected or escaped by that path. The media component then inserts the resulting asset name directly into page/editor link data.

So a filename such as `a[b).png` can still pass upload sanitization while conflicting with Markdown image syntax; there is no Markdown-specific filename validation or warning.

PR #4533, opened 2021-10-02
Prevent malformed paths for pages

### LINKED ISSUE #1368: [Bug] No protection preventing trailing slash on page path
Hello,

Starting a path with a / is not allowed.

image

By error we put one and saved the page. After that the page is no more accessible, and we need to get into the database to change the path manually without beginning /.
This bug can be pretty blocking for people not being technical, and user can put trailing slash easily by error.

Proposed fix:
There should be a test to prevent saving the path if there is a trailing slash.

Another way to change path don't work:
By the way we tried to use the icon to change the path from the admin area, but it open nothing... (is it a feature to come ?). It would have been very usefull in this present case.

image

Thanks.

broxen:
This PR adds some simple form validation to prevent malformed paths in the page properties dialog and shows some error messages if path is malformed (or missing). This has been tested on new pages and editing existing pages.

Closes #1368

Allowed characters: `A-Z` `0-9` `-` `_` _unicode_
## Example Allowed Paths
```
path
folder/path
folder1234/path1234
folder/folder/path
folder/Bushidō-武士道_folder/path
```
## Example Prohibited Paths
```
/path
path/
folder/path#hash
folder/$@#$!@*%&()[]=`~"';:/path
path.ext
```

### DIFF
--- a/client/components/editor/editor-modal-properties.vue
+++ b/client/components/editor/editor-modal-properties.vue
@@ -277,8 +277,8 @@ export default {
       rules: {
           required: value => !!value || 'This field is required.',
           path: value => {
-            const pattern = /^(?![#\/.]).*(?<![#\/.])$/
-            return pattern.test(value) || 'Invalid path. Please ensure it does not begin or end in a slash or hashtag string.'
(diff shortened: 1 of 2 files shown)


broxen:
If this is acceptable, I may be able to apply the same strategy to address _renaming_ files, and possibly uploading them as noted in #1304

broxen:
Ready to go. Thank you for the feedback and review

PR #4663, opened 2021-11-05
Update Dockerfile: set VOLUME to /wiki/data

TomHutter:
By this you can prevent issues like [#5](https://github.com/linuxserver/docker-wikijs/issues/5) with access rights to eg. /wiki/data/cache


NGPixel:
Deferred to next major release as this can potentially cause issues with existing installations / reinstall with scripts that would need to be updated.

PR #4698, opened 2021-11-15
update Dockerfile: set python to python3

delong1998:
When I'm ready to use the dockerfile to build the image
An error has occurred, like this:

ERROR: unable to select packages:
  python (no such package):
    required by: world[python]

And then I go to the address source and I find the python3 package after that, but there's no python2 package before that

### DIFF
--- a/dev/build/Dockerfile
+++ b/dev/build/Dockerfile
@@ -3,7 +3,7 @@
 # ====================
 FROM node:14-alpine AS assets
 
-RUN apk add yarn g++ make python --no-cache
+RUN apk add yarn g++ make python3 --no-cache
 
 WORKDIR /wiki
 


axisofentropy:
I saw this also. I just removed `python` from the list and the build seems to work fine? Does anything in that stage of the build actually require Python? If so, which version of Python?

thewing12:
I also encountered the same yesterday. 

In my case， I changed the default repositories of alpine in the **/dev/build/Dockerfile**, as follows:
```
FROM node:14-alpine AS assets
# change default repositories to v3.11
RUN echo http://mirrors.aliyun.com/alpine/v3.11/main/ > /etc/apk/repositories && \
    echo http://mirrors.aliyun.com/alpine/v3.11/community/ >> /etc/apk/repositories

```
and wiki seems work fine until today.

Hope it can help.

NGPixel:
Closing as the dockerfile was updated to remove python where not necessary.

PR #4714, opened 2021-11-18
Introduce volumes and volumeMounts chart values

jordanjean:
# Description

This PR aims at making possible to add extra volumes and extra volumes mounts in Wiki.js container. It is really useful to add CA certificates in /etc/ssl/certs for instance.

# Test

Deploy the chart with these values and check the mount in the Wiki.js container.

```yaml
[...]

volumeMounts:
  - name: my-volume
    mountPath: /mydir
    readOnly: true
volumes:
  - name: my-volume
    hostPath:
      path: /mydir
```


sajidakram62l (commented):
dev/helm/values.yaml

NGPixel:
Please restore the spaces that were stripped at the end of some lines in the README. These are not mistakes and are necessary to force line returns.

jordanjean:
Hello, sure, I restored the spaces.

NGPixel (approved):

PR #4985, opened 2022-02-08
permit bodyParserLimitSize configuration by ENV variable

matthgyver:
set bodyParserLimitSize by default to 1mb
permit modifications by environment variable


NGPixel:
I modified your PR to use a config value instead of ENV directly, so that it can be used from both the config.yml or ENV.

PR #5053, opened 2022-02-27
Add strings to admin pages

Justman10000:
New strings might need to be added to the translation project...

### DIFF
--- a/client/components/admin/admin-editor.vue
+++ b/client/components/admin/admin-editor.vue
@@ -5,24 +5,30 @@
         .admin-header
           img(src='/_assets/svg/icon-web-design.svg', alt='Editor', style='width: 80px;')
           .admin-header-title
-            .headline.primary--text Editor
-            .subtitle-1.grey--text Configure the content editors #[v-chip(label, color='primary', small).white--text coming soon]
+            .headline.primary--text {{ $t('admin:editor.title') }}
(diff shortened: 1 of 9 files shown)


Justman10000:
@NGPixel 

Why don't you merge this if you're already renaming it?

NGPixel:
There are several issues with your PR:

- Changes are to entire files instead of just the specific lines where strings need to be replaced.
- Many strings don't exist in Lokalise and would need to be created. This would require some work on my side and this is not the priority right now.
- The localization string names don't follow the camelCase format used elsewhere in the project.
- The modification to dev/index.js shouldn't be included in this PR. This bug was fixed recently and removing the call is not the right way to do it.
- You modify admin views that are unfinished and not user-accessible (editors, logging and webhooks). These shouldn't be included in the PR.

Justman10000:
@NGPixel

NGPixel:
Your PR has many issues I listed earlier and your responses to them don't make sense.

Formatting and key naming are wrong. It's not a matter of whether it works or not on your machine.

The main issue still remains that this PR only addresses half of the work needed to support localization. All the strings still need to be created, tagged, described and translated which is a very long task. This is really not the priority right now.

Thanks for the work you've put in but I don't see this PR being merged anytime soon for all the reasons listed above.

Justman10000:
@NGPixel

What if I could help add it?

Justman10000:
Though I thought all the strings would be automatically added to Lokalise...

PR #5114, opened 2022-03-23
Add an example of underline

matejak:
I have found out that `_underline_` resulting in _underline_ is supported, but not documented in the help.

So I have added this to the help page code by copy-pasting and altering corresponding sections, i.e. I have not tested the PR. Please let me know if it is needed, and if there is a corresponding guide that would assist in testing.

### DIFF
--- a/client/components/editor/markdown/help.vue
+++ b/client/components/editor/markdown/help.vue
@@ -31,6 +31,17 @@
                   v-card.editor-markdown-help-result(flat)
                     v-card-text
                       .caption: em Lorem ipsum
+              .body-2.mt-3 Underline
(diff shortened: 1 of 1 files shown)


matejak:
What do you mean by supported but not enabled? When I try the `_underline_` syntax, it works for me in wikijs.

NGPixel:
It works because you enabled underline under admin > rendering.

The default for new installations is off.

matejak:
I see, it was probably an old installation. So if this is an unwanted feature, then it would make sense to close this, and if it is a meaningful feature, this PR should be probably expanded to the state when everything is supported, documented and working at the same time.

Issue #2554, opened 2020-10-08
Multiple auth strategies of same type incorrectly re-use the same instance


XodusMatthewB:
Fresh install of Wiki.js 2.5.144.
Added LDAP / AD strategy for primary domain of acme.com and user login works as expected, without issue.
Added LDAP / AD strategy for secondary domain of this-acme.com and user login works as expected however primary domain authentication now fails.

If I change the strategy authentication order and put primary last, now primary domain authentication works and secondary doesn't.

They work by themselves as the only authentication strategy (besides local), and the last one in the list works but not both at the same time.

Please let me know if any more information is needed.

bvartik-tachyum:
Same issue here. One AD account strategy works, adding a second one breaks the authentication on first one.

chefino:
Just came across the same issue trying to add a second MS AD to authentication strategy.
Once second configuration is added, the first one stops authenticating altogether in my case.

Any workaround for the time being?

XodusMatthewB:
@chefino not that I have found... and I'm sorry you are having the same issue but at the same time glad as maybe now @NGPixel can find some time to take a look at it.  I'm sure he and the team are working on other issues.  Some, if not most, are probably more important however selfishly, this one is my priority :)

FelixSFD:
Same issue on version 2.5.170.

It seems to depend on the position. Only the strategy at the lowest position in the list will be used

![image](https://user-images.githubusercontent.com/2721240/112170079-0f9a0d00-8bf3-11eb-9040-fe3279f5a6d8.png)

bdnettleton:
We're hitting this issue as well with multiple Generic OIDC authentications.

ScottBurfieldMills:
I have experienced this in 2.5.201 using LDAP/Active Directory.

crazyaaron97:
I've the same issue with Version 2.5.201. I checked with tcpdump and all requests go to the secondary ldap connection. No matter which one is choosen.

wellsleep:
Same issue with Wiki.js 2.5.219, two LDAP authentication strategies but only the 2nd works.

rhvitfeldt:
This is still an issue in version 2.5.272. Any news, please? @NGPixel

skytorin:
Same problem. Is a fix expected in version 3?

frantic91:
I fixed it. The main problem was in file:

_server/modules/authentication/ldap/authentication.js_

[...]

PR #5116, opened 2022-03-24
Multi ldpa strategies Issue #2554

frantic91:
I fixed it. The main problem was in file:

_server/modules/authentication/ldap/authentication.js_

Every LDAP strategy has the same name in the strategy list (look at line 13) so only the last added one will work.

How I fixed it:

1.  Open file _server/core/auth.js_
2.  Add `stg.config.key = stg.key;` after line 84.
3. Open file _server/modules/authentication/ldap/authentication.js_
4. Replace line 13 with `passport.use(conf.key,`
5. Open file _server/models/users.js_
6. Replace line 311 with `WIKI.auth.passport.authenticate(selStrategy.key, {`

### DIFF
--- a/server/core/auth.js
+++ b/server/core/auth.js
@@ -82,6 +82,7 @@ module.exports = {
           const strategy = require(`../modules/authentication/${stg.strategyKey}/authentication.js`)
 
           stg.config.callbackURL = `${WIKI.config.host}/login/${stg.key}/callback`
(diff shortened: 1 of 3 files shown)


NGPixel (approved):

yhuaxing:
Item 6 should not be modified. If modified, other user-defined login methods will be affected.

PR #5144, opened 2022-04-05
Fixes #5143

### LINKED ISSUE #5143: graphql update page mutation: Cannot read property 'map' of undefined
### Discussed in https://github.com/requarks/wiki/discussions/3627

<div type='discussions-op-text'>

<sup>Originally posted by **triffid** November 13, 2020</sup>
**Describe the bug**
Getting " Cannot read property 'map' of undefined" for page update mutations

**To Reproduce**
Feeding

	mutation {
		pages {
			update (id:2, content:"This page edited via API") {
				responseResult {
					errorCode,
					message,
					succeeded,
					slug
				}
			}
		}
	}

into https://my-wiki-host/graphql via web browser gives me

	{
		"data": {
			"pages": {
				"update": {
					"responseResult": {
						"errorCode": 1,
						"message": "Cannot read property 'map' of undefined",
						"succeeded": false,
						"slug": "TypeError"
					}
				}
			}
		}
	}

Which seems like either an internal error, or a message that's just as obtuse as the one that appears when I forget to put the authorization token into http headers.

**Expected behavior**
Either page text gets changed, or receive a _meaningful_ error message that suggests the proper way to do it.

 - OS: Docker Container (Linux)
 - Wiki.js version: 2.5.170
 - Database engine: PostgreSQL 11.8

**Additional context**
I'm having a terrible time trying to find any documentation for this API - seems like there's a machine-readable schema but no tutorials or examples anywhere, so I'm left stumbling about in the dark trying to guess what query format I need to do what I want.</div>

splumhoff:
If no tags are submitted e.g. via mutation{pages {update()}} the update will fail. This fixes the issue, same solution as in the pages create method:

https://github.com/requarks/wiki/blob/b5b4b0880ae26f4b137242267b6674b51af8688c/server/models/pages.js#L327

### DIFF
--- a/server/models/pages.js
+++ b/server/models/pages.js
@@ -439,7 +439,9 @@ module.exports = class Page extends Model {
     let page = await WIKI.models.pages.getPageFromDb(ogPage.id)
 
     // -> Save Tags
-    await WIKI.models.tags.associateTags({ tags: opts.tags, page })
+    if (opts.tags && opts.tags.length > 0) {
+      await WIKI.models.tags.associateTags({ tags: opts.tags, page })
+    }
 
     // -> Render page to HTML
     await WIKI.models.pages.renderPage(page)


NGPixel:
You're missing the scenario where you want to remove all tags. Your current code would skip saving because the array length is 0.

PR #5153, opened 2022-04-06
#5139 To fix the image flickering issue, change the v-img tag to img tag.

rajatchauhanyti:
Issue:
There is some weird behavior with v-img tag while display the header logo.
The logo is flickering whenever we are clicking on different file on the side navigation.

Solution:
I have replaced "v-img" tag with "img" tag, The issue is fixed by this simple change.
Additionally i have added the CSS class as "class" attribute to the img tag.

### DIFF
--- a/client/components/common/nav-header.vue
+++ b/client/components/common/nav-header.vue
@@ -21,7 +21,7 @@
       v-flex(xs5, md4)
         v-toolbar.nav-header-inner(color='black', dark, flat, :class='$vuetify.rtl ? `pr-3` : `pl-3`')
           v-avatar(tile, size='34', @click='goHome')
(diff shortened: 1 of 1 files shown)


NGPixel:
This is the not proper way to "fix" it. v-img is not the problem, the default transition is. Also, the classes are meant to be where they were before, not in a `class` attribute, unless they are dynamically generated.

```pug
v-img.org-logo(:src='logoUrl', :transition='false')
```

rajatchauhanyti:
> ```
> v-img.org-logo(:src='logoUrl', :transition='false')
> ```

I am trying with transition='false' as suggested by you but it's not working. Logo is still flickering. Any suggestion...

PR #5157, opened 2022-04-06
Login page logo issue

rajatchauhanyti:
Issue: 
There is black background in the login screen logo and the logo is very small.

Fix
Increased the logo size.
Remove the background color from the image.
Remove the margin-left CSS. 

Please associate this PR with issue #5155


NGPixel:
The logo size and background color are this way to be consistent with the logo on other pages.

Issue #5164, opened 2022-04-07
Administrator general setting is not saving with empty value.


rajatchauhanyti:
### Discussed in https://github.com/requarks/wiki/discussions/5163

<div type='discussions-op-text'>

<sup>Originally posted by **rajatchauhanyti** April  7, 2022</sup>
Steps to reproduce issue:
1. Go to Administrator setting then go to General tab. 
2. Add some value on all the fields in general tab (site URL, title, logo, company name, content license, description, meta robots) and save by clicking the "APPLY" button.
3. Now empty one of the field or else all the possible fields and save by clicking the "APPLY" button.

Observation: 
The changes are not getting saved. The old values are populated once the page is reloaded.

</div>

PR #5165, opened 2022-04-07
#5164 Administrator general setting is not saving with empty value.

rajatchauhanyti:
Solution:
Change the if condition, so instead of checking for the property directly, Checking using "hasOwnProperty" method.

Please associate this PR with issue #5164


NGPixel:
Thanks! Although I'm not sure why you would save any of these settings with an empty value...

rajatchauhanyti:
I don't want to display copyrights string in the footer, so as per implementation if we keep footer copyright and content license as empty. the string will not be displayed.
Also, i want to suggest one more change in the footer, can we hide the entire footer as well, like removing powered by as well from the footer by same configuration as above

NGPixel:
While it can technically be removed, an option to do so will not be added. This is a free, open source software... The least you can do is to have a mention in the footer to support it...

PR #5188, opened 2022-04-14
feat(auth): OAuth2 access_token in GET query string in userInfoURL

triszt4n:
See https://github.com/requarks/wiki/discussions/5187

Small patch for using older OAuth2 authorization servers

### DIFF
--- a/server/modules/authentication/oauth2/authentication.js
+++ b/server/modules/authentication/oauth2/authentication.js
@@ -38,6 +38,8 @@ module.exports = {
 
     client.userProfile = function (accesstoken, done) {
       this._oauth2._useAuthorizationHeaderForGET = true
+      const url = new URL(conf.userInfoURL)
+      if (url.searchParams.has('access_token')) { this._oauth2._useAuthorizationHeaderForGET = false }
       this._oauth2.get(conf.userInfoURL, accesstoken, (err, data) => {
         if (err) {
           return done(err)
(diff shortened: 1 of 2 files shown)


triszt4n:
I'm still thinking about adding only a prop in the definition.yml instead. My solution might create backward-compatibility issues (well, only if someone relies on using an `access_token` searchParam already for other purposes in their userInfoURL).

NGPixel:
This should indeed be optional and off by default.

triszt4n:
@NGPixel Applied your suggested modifications, tested it manually with fresh database with my oauth2 config, worked very well.

server/modules/authentication/oauth2/authentication.js
```diff
@@ -37,7 +37,13 @@ module.exports = {
     })
 
     client.userProfile = function (accesstoken, done) {
-      this._oauth2._useAuthorizationHeaderForGET = true
+      const url = new URL(conf.userInfoURL)
+      if (conf.useAuthorizationHeader) {
+        this._oauth2._useAuthorizationHeaderForGET = true
+      } else {
+        this._oauth2._useAuthorizationHeaderForGET = false
+        url.searchParams.set('access_token', accesstoken)
```
NGPixel, line 45: You're creating an `url` variable, set a searchParams but then it's never used again? Am I missing something?

Also the current default would cause existing installations to suddenly use the opposite behavior. Either set the useAuthorizationHeader default as true or reverse the logic...

NGPixel (changes requested):

server/modules/authentication/oauth2/authentication.js
triszt4n: Completely forgot about the url variable. Will reverse the logic. Thx for the review gonna mend the things and test them again.

NGPixel (approved):

PR #5192, opened 2022-04-18
fix(logout): Invalidate user jwt token after logout

myml:


Vincent-Nikkelen-mkb:
Hi,
I am not a node.js developer, but after a quick look, am I correct that you try to store the token without having established it is valid?

In other words, is there an attack vector where an attacker can call /logout a lot which will cause a memory problem?

Side-note: Personally I like to avoid having state in the back-end, and a list of logout tokens is state for me. The JWT token responsibility should be 100% at the client. This solution will not scale when there are multiple instances of the back-end as the cache isn't shared.

NGPixel:
This implementation indeed doesn't scale and is prone to attacks.

The default expiration for JWT is 30 minutes. If you need to forcefully logout someone immediately, you can regenerate the certificates under Admin > Utilities > Authentication.

myml:
> Hi, I am not a node.js developer, but after a quick look, am I correct that you try to store the token without having established it is valid?
> 
> In other words, is there an attack vector where an attacker can call /logout a lot which will cause a memory problem?
> 
> Side-note: Personally I like to avoid having state in the back-end, and a list of logout tokens is state for me. The JWT token responsibility should be 100% at the client. This solution will not scale when there are multiple instances of the back-end as the cache isn't shared.

Thank you for the reminder! 
The security of token from company to company, this pr is not perfect solution, I close if for now.

PR #5197, opened 2022-04-19
feat: password encrypted transmission

myml:
The password is encrypted in the browser, then send to the backend and
then decrypeted.

### DIFF
--- a/client/components/admin/admin-users-create.vue
+++ b/client/components/admin/admin-users-create.vue
@@ -92,6 +92,7 @@
 import _ from 'lodash'
 import validate from 'validate.js'
 import gql from 'graphql-tag'
+import crypto from 'crypto-browserify'
 
 import createUserMutation from 'gql/admin/users/users-mutation-create.gql'
 import groupsQuery from 'gql/admin/users/users-query-groups.gql'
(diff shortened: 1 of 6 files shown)


NGPixel:
I'm not sure what this achieves exactly? The connection between the client and the server is already encrypted by the underlying HTTPS protocol...

myml:
Thank you for attention. HTTPS is great, But the real world has CDN, Cloud Service Gateway... From the browser to the wikijs service is not always https and people like to use the same password for multiple services.

References:
1. [NSA infiltrates links to Yahoo, Google data centers worldwide, Snowden documents say](https://www.washingtonpost.com/world/national-security/nsa-infiltrates-links-to-yahoo-google-data-centers-worldwide-snowden-documents-say/2013/10/30/e51d661e-4166-11e3-8b74-d89d714ca4dd_story.html?itid=lk_inline_manual_1)
2. [GitHub Accidentally Recorded Some Plaintext Passwords in Its Internal Logs](https://www.bleepingcomputer.com/news/security/github-accidentally-recorded-some-plaintext-passwords-in-its-internal-logs/)

NGPixel:
Sorry I don't agree, HTTPS is the only proper solution, not encrypting passwords client-side. It doesn't protect against man in the middle attacks either (the value could be replaced with another encrypted secret).

It only adds unnecessary complexity to the code while providing near-zero security improvements.

NGPixel:
Sorry I don't agree, HTTPS is the only proper solution, not encrypting passwords client-side. It doesn't protect against man in the middle attacks either (the value could be replaced with another encrypted secret).

It only adds unnecessary complexity to the code while providing near-zero security improvements.

PR #5207, opened 2022-04-21
fix: docker build fail

myml:
cpu-features build fail

[github action 'Build and push Docker images' stage log error](https://github.com/requarks/wiki/runs/6050848396?check_suite_focus=true#step:7:339)

image


NGPixel (approved):

NGPixel:
Thanks, although that module is not used at all during this build stage.

PR #5215, opened 2022-04-25
fix: Show undefined author in git commit when delete page

myml:
image

### DIFF
--- a/server/models/pages.js
+++ b/server/models/pages.js
@@ -787,15 +787,7 @@ module.exports = class Page extends Model {
    * @returns {Promise} Promise with no value
    */
   static async deletePage(opts) {
-    let page
-    if (_.has(opts, 'id')) {
-      page = await WIKI.models.pages.query().findById(opts.id)
-    } else {
-      page = await WIKI.models.pages.query().findOne({
-        path: opts.path,
-        localeCode: opts.locale
-      })
-    }
+    const page = await WIKI.models.pages.getPageFromDb(_.has(opts, 'id') ? opts.id : opts);
     if (!page) {
       throw new WIKI.Error.PageNotFound()
     }


server/models/pages.js
```diff
@@ -789,12 +789,9 @@ module.exports = class Page extends Model {
...
     if (_.has(opts, 'id')) {
-      page = await WIKI.models.pages.query().findById(opts.id)
+      page = await WIKI.models.pages.getPageFromDb(opts.id) 
     } else {
-      page = await WIKI.models.pages.query().findOne({
-        path: opts.path,
-        localeCode: opts.locale
-      })
+      page = await WIKI.models.pages.getPageFromDb(opts)
     }
```
NGPixel, line 795: Could simplify to just:

```js
const page = await WIKI.models.pages.getPageFromDb(_.has(opts, 'id') ? opts.id : opts)
```

NGPixel (changes requested):

server/models/pages.js
```diff
@@ -787,15 +787,7 @@ module.exports = class Page extends Model {
...
-    let page
-    if (_.has(opts, 'id')) {
-      page = await WIKI.models.pages.query().findById(opts.id)
-    } else {
-      page = await WIKI.models.pages.query().findOne({
-        path: opts.path,
-        localeCode: opts.locale
-      })
-    }
+    const page = await WIKI.models.pages.getPageFromDb(_.has(opts, "id") ? opts.id : opts);
```
NGPixel, line 790: Minor thing but use single quotes around `id`, for code style consistency.

NGPixel (changes requested):

NGPixel (approved):

PR #5220, opened 2022-04-25
feat: Added lightbox/modal support for image preview

TarikTz:
Add a simple "modal" class to images from the assets dialog and enable a modal window for preview.  Clicking on the modal window will close it.

 

Example for a thumbnail preview with a clickable image

```
[![title](/example.jpg =300x)](/example.jpg){.modal}
```

or regular inserted image
```
![example.jpg](/example.jpg){.align-center}{.modal}
```

### DIFF
--- a/client/components/editor/editor-code.vue
+++ b/client/components/editor/editor-code.vue
@@ -233,6 +233,9 @@ export default {
           if (opts.align && opts.align !== '') {
             img += ` class="align-${opts.align}"`
           }
+          if (opts.className && opts.className !== '') {
+            img += ` class="align-${opts.className}"`
+          }
           img += ` />`
           this.insertAtCursor({
             content: img
--- a/client/components/editor/editor-markdown.vue
+++ b/client/components/editor/editor-markdown.vue
@@ -849,6 +849,9 @@ export default {
           if (opts.align && opts.align !== '') {
             img += `{.align-${opts.align}}`
           }
+          if (opts.className && opts.className !== '') {
+            img += `{.${opts.className}}`
+          }
           this.insertAtCursor({
             content: img
           })
--- a/client/helpers/modal.js
+++ b/client/helpers/modal.js
@@ -1,0 +1,36 @@
+window.onload = function() {
+  const lightboxImages = window.document.querySelectorAll('.modal')
+
+  lightboxImages.forEach(img => {
+    img.addEventListener('click', e => {
+      e.preventDefault()
+
+      const imageContainer = document.createElement('div')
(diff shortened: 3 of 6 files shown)


client/components/editor/editor-code.vue
```diff
@@ -233,6 +233,9 @@ export default {
           if (opts.align && opts.align !== '') {
             img += ` class="align-${opts.align}"`
           }
+          if (opts.className && opts.className !== '') {
+            img += ` class="align-${opts.className}"`
+          }
```
NGPixel, line 238: This would add a duplicate class if align is also set.

client/components/editor/editor-markdown.vue
```diff
@@ -849,6 +849,9 @@ export default {
           if (opts.align && opts.align !== '') {
             img += `{.align-${opts.align}}`
           }
+          if (opts.className && opts.className !== '') {
+            img += `{.${opts.className}}`
+          }
```
NGPixel, line 854: Same issue as above.

NGPixel (changes requested):

client/components/editor/editor-code.vue
TarikTz: image

It will render something like this if both options are added

client/helpers/modal.js
```diff
@@ -0,0 +1,36 @@
+window.onload = function() {
+  const lightboxImages = window.document.querySelectorAll('.modal')
```
karladler, line 2: you probably don't want to have a lightbox in any case where somebody uses a modal class. Since `modal` might be also used for other purposes, I would at least narrow it to img tags, but better refactor it to something more specific like `lightbox-modal`.

```suggestion
  const lightboxImages = window.document.querySelectorAll('img.modal')
```

client/components/editor/editor-code.vue
karladler: I'm not sure, why we can't use `:class="{'classname': condition, ... }"` syntax here, but its probably a pug thing.
Maybe using function from [classNames](https://github.com/JedWatson/classnames) module might be a solution here. 

Anyways the logic is not too simple and might be moved to it's own function.

`
function getClassProp(opts) {
    if (!opts || (!opts.align && !opts.className)) {
        return '';
    }

    const classes = [];

    if (opts.align) {
        classes.push(`align-${opts.align}`);
    }

    if (opts.className) {
        classes.push(opts.className);
    }

    const classProp = classes.length > 0 ? ` class="${classes.join(' ')}"` : '';

    return classProp;
}
`

PR #5221, opened 2022-04-25
feat: check create folder permissions

lucas-it:
I found an issue (I think it's an issue) recently in the asset management.
When you have the write:assets permission you can create folders anywhere you want even with page rules configured like this :
image

With my fix, when you try to create a folder which is not in the path authorized by the page rules you have this error :
image
And you can create folders when the folder path matches the page rule. In my case, you can create the folder test-asset and subfolders in it.
image

### DIFF
--- a/server/helpers/error.js
+++ b/server/helpers/error.js
@@ -1,6 +1,10 @@
 const CustomError = require('custom-error-instance')
 
 module.exports = {
+  AssetCreateFolderForbidden: CustomError('AssetCreateFolderForbidden', {
+    message: 'You are not authorized to create this folder.',
+    code: 2010
+  }),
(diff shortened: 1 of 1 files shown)


NGPixel:
One potential issue I see is if you only have write access to a specific path. You wouldn't be able to create the folders leading to that path because you don't have write access to the parent paths.

The current asset system is a bit confusing at the moment tbh, which I hope to revamp in v3.

lucas-it:
> One potential issue I see is if you only have write access to a specific path. You wouldn't be able to create the folders leading to that path because you don't have write access to the parent paths.
> 
> The current asset system is a bit confusing at the moment tbh, which I hope to revamp in v3.

I'm currently using Wiki.js and I need this fix. What I need to do ? Juste wait for the v3 or can I improve my code ?

PR #5226, opened 2022-04-26
Solving issue #5200

EyalSeagull:
Comments stoped working in version 2.5.278 (#5200). This patch returns comments funcionallity. Tested with Default Comments, Disqus Comments, both with the "Do not allow comments" option turned on and off.

### DIFF
--- a/server/controllers/common.js
+++ b/server/controllers/common.js
@@ -525,8 +525,8 @@ router.get('/*', async (req, res, next) => {
           }
 
           // -> Inject comments variables
-          const shouldDisplayComments = WIKI.config.features.featurePageComments;
-          let commentTmpl = shouldDisplayComments ? {
+          const shouldDisplayComments = WIKI.config.features.featurePageComments
+          const commentTmpl = shouldDisplayComments ? {
             codeTemplate: WIKI.data.commentProvider.codeTemplate,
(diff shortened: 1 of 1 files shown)


server/controllers/common.js
```diff
@@ -525,8 +525,8 @@ router.get('/*', async (req, res, next) => {
           }
 
           // -> Inject comments variables
-          const shouldDisplayComments = WIKI.config.features.featurePageComments && WIKI.data.commentProvider.codeTemplate
-          const commentTmpl = shouldDisplayComments ? {
+          const shouldDisplayComments = WIKI.config.features.featurePageComments;
+          let commentTmpl = shouldDisplayComments ? {
```
NGPixel, line 529: No semicolons.

`const` is the proper definition here, not `let`

NGPixel (changes requested):

server/controllers/common.js
EyalSeagull: I see, however when we take off the semicolon and use `const`, it stop working.

NGPixel:
Fixed by 13890a92ab354d35923cea904116575029f78d45

Looking more in depth at the issue, your PR was more of a workaround than a proper solution. The fix above correctly skips the template injection when the default provider is selected.

PR #5233, opened 2022-04-28
Updated MDI Font to allow usage of way newer icons

SeaLife:
Hey,

i updated the `@mdi/font` dependency to the newest which should break nothing but allow the usage of way newer icons :)

I wanted to use `mdi-airplane-marker` which was not available in 5.8.x

image

So, here i go, updating, tested it, seems to be working fine :)

### DIFF
--- a/dev/helm/templates/deployment.yaml
+++ b/dev/helm/templates/deployment.yaml
@@ -51,7 +51,7 @@ spec:
             {{- end }}
             - name: DB_TYPE
               value: postgres
-            {{- if (.Values.externalPostgresql).databaseURL }}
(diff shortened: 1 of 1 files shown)


NGPixel:
There're about 100 icons that got either removed or renamed going from 5.x to 6.x, so this would be a breaking change.

SeaLife:
Thats true https://github.com/Templarian/MaterialDesign/issues/5409
Is there a way to get this on to the main branch anyway?
There is no way we could use both versions is there?

funjoker:
how about adding a setting to select the version?

PR #5285, opened 2022-05-12
fix: changed spelling of Kiev to Kyiv

crambo:
Kiev is a Russian style of spelling. The proper and government official way is Kyv.
More details is available [there](https://en.wikipedia.org/wiki/KyivNotKiev).


nasiralamreeki:
This pull request should be rejected as it's actually Київ in Ukrainian and the [Ukranian Foreign Ministry prefers the transliteration of "Kyiv"](https://en.m.wikipedia.org/wiki/KyivNotKiev) and the [ISO standard calls for "Kyiv" too](https://www.iso.org/obp/ui/#iso:code:3166:UA) and that's what's followed in software development.

nasiralamreeki:
What the Ukrainian Foreign Ministry prefers shouldn't influence software development rather we should look to the ISO standard which does use "Kyiv"

NGPixel (approved):

PR #5286, opened 2022-05-13
fix: current directory link without join locale

myml:


NGPixel (approved):

peteruithoven:
Is there any change this change could be causing the Site Tree issue described in the following discussion?
https://github.com/requarks/wiki/discussions/5295

PR #5294, opened 2022-05-16
feat(admin): export tool for full migration / backup

NGPixel:


dberardo-com:
Is it possible to use this export / backup feature from CLI (example using it in a cronjob)?

PR #5314, opened 2022-05-21
Enable youtube and vimeo media players

peteruithoven:
Currently inserted media is sanitized away. 
When inserting media using the visual editor and html sanitation is turned off only an `oembed` tag is added. 

Based on an [existing workaround](https://github.com/requarks/wiki/discussions/5043#discussioncomment-2787614) I'm letting the `html-mediaplayers` module replace the `oembed` with an youtube or vimeo ready `iframe` when it's a youtube or vimeo video.  
I've also added a few common iframe attributes to the html-security module, like `allowfullscreen`. 

This change will still require people to enable "Allow iframes", which might be a bit confusing.

This is probably far from the perfect solution but let's get the ball rolling, please let me know what to improve.

### DIFF
--- a/server/modules/rendering/html-mediaplayers/renderer.js
+++ b/server/modules/rendering/html-mediaplayers/renderer.js
@@ -1,5 +1,26 @@
+const rxYoutube = /^.*(?:(?:youtu\.be\/|v\/|vi\/|u\/\w\/|embed\/)|(?:(?:watch)?\?v(?:i)?=|&v(?:i)?=))([^#&?]*).*/
+const rxVimeo = /^.*(vimeo\.com\/)((channels\/[A-z]+\/)|(groups\/[A-z]+\/videos\/))?([0-9]+)/
+
 module.exports = {
   init($, config) {
-
+    $('oembed').each((i, elm) => {
+      const url = $(elm).attr('url')
+      const ytMatch = url.match(rxYoutube)
+      let src
+      if (ytMatch) {
+        src = `https://www.youtube.com/embed/${ytMatch[1]}`
+      } else {
+        const vmMatch = url.match(rxVimeo)
+        if (vmMatch) {
+          src = `https://player.vimeo.com/video/${vmMatch[5]}`
+        } else {
+          return
+        }
+      }
+      const newElm = $(`<div class="video-responsive">
+        <iframe type="text/html" width="640" height="360" src="${src}" frameborder="0" allowfullscreen></iframe>
+      </div>`)
+      $(elm).parent().replaceWith(newElm)
+    })
   }
 }
--- a/server/modules/rendering/html-security/renderer.js
+++ b/server/modules/rendering/html-security/renderer.js
@@ -29,7 +29,7 @@ module.exports = {
 
       if (config.allowIFrames) {
         allowedTags.push('iframe')
-        allowedAttrs.push('allow')
+        allowedAttrs.push('allow', 'frameborder', 'allowfullscreen')
       }
 
       input = DOMPurify.sanitize(input, {


peteruithoven:
Reading https://oembed.com/ should we do a request to for example youtube to get the right html to inject or info to generate the right embed? 

For example:
https://www.youtube.com/oembed?url=https://www.youtube.com/watch?v=m9QU69958JU&format=json
Gives me:
``` json
{
  "title": "Blade Runner 2049 | After Dark - Mr.Kitty",
  "author_name": "Yudong Wang",
  "author_url": "https://www.youtube.com/channel/UCRHMVllLrevpEEKVNZmXHgw",
  "type": "video",
  "height": 113,
  "width": 200,
  "version": "1.0",
  "provider_name": "YouTube",
  "provider_url": "https://www.youtube.com/",
  "thumbnail_height": 360,
  "thumbnail_width": 480,
  "thumbnail_url": "https://i.ytimg.com/vi/m9QU69958JU/hqdefault.jpg",
  "html": "\u003ciframe width=\u0022200\u0022 height=\u0022113\u0022 src=\u0022https://www.youtube.com/embed/m9QU69958JU?feature=oembed\u0022 frameborder=\u00220\u0022 allow=\u0022accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture\u0022 allowfullscreen\u003e\u003c/iframe\u003e"
}
```

This does mean we'd need to trust youtube's html or use the info to build our own html.

peteruithoven:
Actually, I realize that the CKEditor already goes through this process, would be ideal if we could reuse that logic? Would also make the rendered result match with the preview.

This clarifies the CKeditor's oembed output:
https://ckeditor.com/docs/ckeditor5/latest/features/media-embed.html#data-output-format

Looking at the vega branch we'll continue using CKEditor. 

This documentation seems interesting: 
https://ckeditor.com/docs/ckeditor5/latest/features/media-embed.html#displaying-embedded-media-on-your-website

If we're going with a front-end approach this would require that the oembed is no longer filtered out by the `html-security` module.

peteruithoven:
For markdown, where we use markdown-it, we'd also need a solution. There for example is [markdown-it-oembed](https://www.npmjs.com/package/markdown-it-oembed), but that also includes the list of providers, meaning another part to keep in sync. The library's last release was more than a year ago, which isn't a good sign. 

If we do the oembed > preview ourselves using something like [oembed-spec](https://moiva.io/?npm=iframely+oembed-parser+oembed-spec) in the `html-mediaplayers` module, people could add something like this to their markdown:
``` html
<oembed url="https://www.youtube.com/watch?v=m9QU69958JU" />
```

Ideally we add some plugin that converts something like:
``` markdown
![Getting Started: Sktch.io](https://www.youtube.com/watch?v=CMfBj0U221k)
```
Into an oembed html element.

And ideally we have some kind of insert media UI so people don't have to remember the syntax.

peteruithoven:
I'd love some input on the general approach. I think we'd want the solution to at least potentially work for all the editors? 
This means we sadly can't simply go with CKEditor's `previewsInData` feature or [markdown-it-oembed](https://www.npmjs.com/package/markdown-it-oembed). 

What do we think about adding a library like [oembed-spec](https://www.npmjs.com/package/oembed-spec) ([alternatives](
https://moiva.io/?npm=iframely+oembed-parser+oembed-spec))? This externalizes the actual oembed to iframe conversion logic to a library and using a widely used [providers.json file](https://oembed.com/providers.json) from oembed.com. 
But do we even want to support all of those? Seeing we need a bit of CSS to make them responsive, there are a lot of edge cases. 

People would still need to enable iframes in the rendering > security config, which isn't exactly intuitive. It's also not ideal that we always show the Insert Media button in the visual editor since it will never work when allow iframes is disabled. People might not understand that embedding video's requires iframes? What if we add some kind of Enable embedding switch that will show/hide the insert media buttons and controls whether iframes can be added? A description could then explains that this enables iframes and some of the security risks.

peteruithoven:
Possibly interesting is how Wordpress handles this. 

General docs: https://wordpress.org/support/article/embeds/ 

They seem to have a hardcoded list of regular expressions with a matching oembed url:
https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-oembed.php#L53-L107 

How they fetch the oembed data:
https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-oembed.php#L525-L549

Some exceptions of how they handle the result:
https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-oembed.php#L685-L713
(For video they just take the `html` for example)

They have added an API that enables people to add or remove providers.

RHeynsZa:
Can we please merge this?

maxp-hover:
Even with HTML sanitization and iframe stripping turned off, the iframe still doesn't show up on my HTML page. Anyone have any idea what's the cause here?

PR #5318, opened 2022-05-23
feat: XML Sitemap

patrickjvieira:
XML is available in /sitemap.xml if enabled in SEO settings.
https://feedback.js.wiki/wiki/p/automatically-generate-a-sitemap-for-the-wiki

### DIFF
--- a/server/controllers/common.js
+++ b/server/controllers/common.js
@@ -43,7 +43,7 @@ router.get('/sitemap.xml', async function (req, res) {
       // Fix checkAccess attr
       page.locale = page.localeCode
 
-      if (!WIKI.auth.checkAccess(req.user, ['read:pages'], page)) return
+      if (!WIKI.auth.checkAccess(WIKI.auth.guest, ['read:pages'], page)) return
 
       const date = moment(page.updatedAt).format()
       xmlTree.push(`<url><loc>${host}/${page.localeCode}/${page.path}</loc><lastmod>${date}</lastmod></url>`)
--- a/server/controllers/common.js
+++ b/server/controllers/common.js
@@ -22,6 +22,20 @@ router.get('/robots.txt', (req, res, next) => {
 })
 
 /**
+ * Sitemap.xml
+ */
+router.get('/sitemap.xml', async function (req, res) {
+  !WIKI.config.seo.sitemap && res.status(404).end()
+
+  try {
+    const xml = await WIKI.models.pages.buildXMLSitemap()
+    res.header('Content-Type', 'application/xml').send(xml)
+  } catch (e) {
+    res.status(404).end()
+  }
+})
+
+/**
  * Health Endpoint
  */
 router.get('/healthz', (req, res, next) => {
--- a/server/controllers/common.js
+++ b/server/controllers/common.js
@@ -25,11 +25,34 @@ router.get('/robots.txt', (req, res, next) => {
  * Sitemap.xml
  */
 router.get('/sitemap.xml', async function (req, res) {
+  // End if sitemap is enabled
   !WIKI.config.seo.sitemap && res.status(404).end()
 
+  const host = WIKI.config.host
+  const xmlTree = [
+    '<?xml version="1.0" encoding="UTF-8"?>',
+    '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">'
+  ]
+
   try {
-    const xml = await WIKI.models.pages.buildXMLSitemap()
-    res.header('Content-Type', 'application/xml').send(xml)
+    const pages = await WIKI.models.pages.query()
+      .select(['path', 'localeCode', 'updatedAt'])
+      .where('isPublished', '=', true)
+      .where('isPrivate', '=', false)
+
+    pages.forEach(page => {
+      // Fix checkAccess attr
+      page.locale = page.localeCode
+
+      if (!WIKI.auth.checkAccess(req.user, ['read:pages'], page)) return
+
+      const date = moment(page.updatedAt).format()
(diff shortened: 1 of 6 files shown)


server/models/pages.js
```diff
@@ -926,6 +927,29 @@ module.exports = class Page extends Model {
...
+    const xmlTree = [
+      '<?xml version="1.0" encoding="UTF-8"?>',
+      '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">'
+    ]
+
+    const pages = await WIKI.models.pages.query().select(['path', 'updatedAt']).where('isPublished', '=', true)
+    const base = `${WIKI.config.host}/${WIKI.config.lang.code}`
+
+    pages.forEach(page => {
+      const date = moment(page.updatedAt).format('YYYY-MM-DDThh:mm:ssZ')
```
NGPixel, line 944: No need to load moment to output an ISO date string. Date has a native [toISOString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) method.

server/models/pages.js
```diff
@@ -926,6 +927,29 @@ module.exports = class Page extends Model {
...
+   *
+   * @returns
+   */
+  static async buildXMLSitemap() {
+    const xmlTree = [
+      '<?xml version="1.0" encoding="UTF-8"?>',
+      '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">'
+    ]
+
+    const pages = await WIKI.models.pages.query().select(['path', 'updatedAt']).where('isPublished', '=', true)
```
NGPixel, line 940: You load all pages regardless of their permissions. Pages should match what the Guest group can see.

server/models/pages.js
```diff
@@ -926,6 +927,29 @@ module.exports = class Page extends Model {
     return rebuildJob.finished
   }
 
+  /**
+   *
+   * @returns
+   */
+  static async buildXMLSitemap() {
```
NGPixel, line 934: Should be part of the controller directly as it's a single purpose method.

NGPixel (changes requested):

server/models/pages.js
patrickjvieira: I lose timezone with native method, any suggestion?

server/models/pages.js
patrickjvieira: Gotcha!

server/models/pages.js
patrickjvieira: Gotcha!

server/models/pages.js
patrickjvieira: toISOString() has no timezone offset, any suggestion?

server/controllers/common.js
```diff
@@ -21,6 +21,42 @@ router.get('/robots.txt', (req, res, next) => {
...
+    const pages = await WIKI.models.pages.query()
+      .select(['path', 'localeCode', 'updatedAt'])
+      .where('isPublished', '=', true)
+      .where('isPrivate', '=', false)
+
+    pages.forEach(page => {
+      // Fix checkAccess attr
+      page.locale = page.localeCode
+
+      if (!WIKI.auth.checkAccess(req.user, ['read:pages'], page)) return
```
NGPixel, line 46: Use `WIKI.auth.guest` instead of `req.user` to provide a consistent output to all users.

server/models/pages.js
NGPixel: Indeed. Leave moment then. It will be replaced by luxon in v3.

server/controllers/common.js
patrickjvieira: Indeed, now is no difference between logged and unlogged users.

NGPixel:
You should also save the xml output to disk to avoid querying the DB on every page hit. Something like...

```js
const cachePath = path.resolve(WIKI.ROOTPATH, WIKI.config.dataPath, `cache/sitemap.xml`)

try {
  const sitemap = await fs.readFile(cachePath, 'utf8')
  res.header('Content-Type', 'application/xml').send(sitemap)
} catch (err) {
  if (err.code === 'ENOENT') {
    
    // CODE TO GENERATE SITEMAP HERE

    await fs.outputFile(cachePath, sitemap)
    return sitemap
  } else {
    WIKI.logger.error(err)
    throw err
  }
}
```

patrickjvieira:
> You should also save the xml output to disk to avoid querying the DB on every page hit. Something like...
> 
> ```js
> const cachePath = path.resolve(WIKI.ROOTPATH, WIKI.config.dataPath, `cache/sitemap.xml`)
> 
> try {
>   const sitemap = await fs.readFile(cachePath, 'utf8')
>   res.header('Content-Type', 'application/xml').send(sitemap)
> } catch (err) {
>   if (err.code === 'ENOENT') {
>     
>     // CODE TO GENERATE SITEMAP HERE
> 
>     await fs.outputFile(cachePath, sitemap)
>     return sitemap
>   } else {
>     WIKI.logger.error(err)
>     throw err
>   }
> }
> ```

How long does the cache persist? Isn't better to generate the Sitemap after saving some page?

patrickjvieira:
How long does the cache persist? Isn't better to generate the Sitemap after saving some page?

NGPixel:
The cache should indeed be cleared when making changes to pages. Add a line to:

https://github.com/requarks/wiki/blob/main/server/models/pages.js#L1112

to remove the `sitemap.xml` file. This method gets called when a page is modified / moved / deleted. It's not called when creating a new page though, so you would need to delete the sitemap in the createPage() method as well.

myml:
Nice!

mrbrianevans:
I'm happy to work on this, I can implement what @NGPixel has suggested

PR #5326, opened 2022-05-25
fix: try to find home, redirect if found (#4500)

kesuskim:
This PR solves issues of #4500 by trying to find home if page not found, and redirect if home is found.

### DIFF
--- a/server/controllers/common.js
+++ b/server/controllers/common.js
@@ -433,6 +433,18 @@ router.get('/*', async (req, res, next) => {
         userId: req.user.id,
         isPrivate: false
       })
+      // -> Try to find home if page not found
+      if (!page) {
+        // -> If page found
+        if (await WIKI.models.pages.getPage({
+          path: `${pageArgs.path}/home`,
+          locale: pageArgs.locale,
+          userId: req.user.id,
+          isPrivate: false
+        })) {
+          return res.redirect(`${pageArgs.path}/home`)
+        }
+      }
       pageArgs.tags = _.get(page, 'tags', [])
 
       // -> Effective Permissions


NGPixel:
I'm not sure I understand what problem this PR is trying to fix? The linked issue is very vague as well.

kesuskim:
Primary problem of this, I believe, is because confusion of being root page named 'home', that for example if you are trying to locate root page manually by entering URL in browser without `home` it leads to create new page.

And my cause of this PR was using breadcrumbs of page; if there are multiple level of pages, and if I click one of breadcrumbs between them, it should lead to 'home' page, but does not which leads me to create new page, which is very confusing because I already created 'home' page for that level, but breadcrumbs does not tell me to go home.

At first, I tried to fix client side (https://github.com/requarks/wiki/blob/main/client/themes/default/components/page.vue#L489-L497), but as there is no information of among many nested pages which level does have home page, so I think this should be fixed from server, just first try to find page, if it cannot find the page, try to find home again, if not, goes what it was meant to be is enough.

NGPixel:
I think you're confusing what corresponds to a "folder" page. There's only a single `/home` which is an alias for the page at path `/`. There's no such thing as home for nested levels.

If you have a page at path `/foo/bar`, then you should create a new page at `/foo`.

For example, if you go to https://docs.requarks.io/editors/markdown and click on the "editors" link in the breadcrumbs, it will go to the page I created at path `/editors`

kesuskim:
OK, then this is obviously misleading 😕 maybe #4500 also can be closed too (I think?)

NGPixel:
I'll add some more info to the docs to make it clearer.

kesuskim:
I think 'folder page' concept should be explicitly explained 😃

kesuskim:
Also I think main reason for confusion is there is no explicit way to create folder. It is page if there is only one page in that level, and it becomes directory if there are more than one page. 😖

NGPixel:
See https://docs.requarks.io/en/guide/structure#nesting

PR #5330, opened 2022-05-25
Load self-signed CA from file as described in the helm chart

monneyboi:
The helm chart supports [`postgresql.ca`](https://github.com/requarks/wiki/blob/main/dev/helm/values.yaml#L120) for setting self-signed CA's for postgres connections, though when trying this i noticed that the servers database logic handles this differently.

It would load a CA from a concatenated string, which i also tried, but this also didn't work, as it would still set `rejectUnauthorized: true`, thereby disalllowing self-signed CA certificates. 

If you're CA would be signed by a root CA, there would be no need to set a CA in the first place, so i guess this logic never worked?

With this pull request, the `DB_SSL_CA` logic now corresponds to the descriptions in the helm chart.

### DIFF
--- a/server/core/db.js
+++ b/server/core/db.js
@@ -60,17 +60,13 @@ module.exports = {
       sslOptions = true
     }
 
-    // Handle inline SSL CA Certificate mode
+    // Handle self-signed CA file
+    // https://node-postgres.com/features/ssl
     if (!_.isEmpty(process.env.DB_SSL_CA)) {
-      const chunks = []
-      for (let i = 0, charsLength = process.env.DB_SSL_CA.length; i < charsLength; i += 64) {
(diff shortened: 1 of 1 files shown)


NGPixel:
Your PR would break the existing behavior (concatenated string) and switching to loading a file from disk. That env variable isn't only used in Helm... it's used for simple docker containers as well (or even native environments).

I agree with you concerning `rejectUnauthorized: true`, which should be configurable. I've read scenarios where you expect it to be true and still use a self-signed certificate so both settings are not mutually exclusive.

A better solution would be to either check if the variable is a path and handle the parsing differently if so, or have a different variable altogether specifically for providing a CA path (e.g. `DB_SSL_CA_PATH`).

monneyboi:
I've now added back the concatenated string logic and made `rejectUnauthorized` configurable through environment variable.

PR #5333, opened 2022-05-25
Fix sidebar root title by not overiding it when translation isn't yet available

peteruithoven:
Fixes: https://github.com/requarks/wiki/discussions/4888

Before:
Screenshot before@2x
After:
Screenshot after@2x

I'm not sure why but the root parent's title is being overriden with a translatable title. But this happens in the mount handler when the translation isn't yet available, causing it to fallback to the text "sidebar.root".


NGPixel:
Your PR completely removes translation support. You should instead add a fallback value:

`this.$t('common:sidebar.root', '/ (root)')`

NGPixel:
Your PR completely removes translation support. You should instead add a fallback value:

`this.$t('common:sidebar.root', '/ (root)')`

peteruithoven:
I'll have another look at making the title adapt to translation updates, but I don't think what you're suggesting will do anything. 
In the `data` the `currentParent` already gets that `'/ (root)'` as a default. I've seen that during the mount event the translation isn't retrieved yet.

peteruithoven:
What do you think about this approach?

peteruithoven:
Could you have another look?

kimegede:
Looking for this as well :)

PR #5334, opened 2022-05-26
fix: don't push files to git if ignored

Fireant456:
This was an edge case we ran into wanting to keep internal documentation off our synced git storage. This change allows us to maintain our own gitignore file to exclude files from git storage and still sync them to other storage backends.


NGPixel:
Interesting.

Just one small change, rename variable `gitPath` to `gitFilePath` so that it's more obvious that we're talking about a file path and not the location of the git binary.

Fireant456:
> `gitFilePath`

Completed

NGPixel (approved):

PR #5350, opened 2022-05-31
Fix page cursor and pagination

zemzot:
### Page Cursor

fix: page cursor position stays the same between page selection
Given a query with multiple result pages, if the cursor is moved to a result entry, then the user selects a different page, the cursor remains on the previous position.
wiki js-fix-page-cursor

### Pagination

fix: search results not displaying on first page
Given a query with multiple result pages, if a certain page is selected, then another query (with the same or more result pages as the previous entry) is searched, the results will display on the previously selected page. 
wiki js-pagination


client/components/common/search-results.vue
```diff
@@ -84,6 +84,7 @@ export default {
     searchRestrictLocale: sync('site/searchRestrictLocale'),
     searchRestrictPath: sync('site/searchRestrictPath'),
     results() {
+      this.cursor = 0
```
NGPixel, line 87: Don't modify the state from within a computed getter. Use a watcher instead, e.g.:

```js
watch: {
  results () {
    this.cursor = 0
  }
}
```

NGPixel (changes requested):

client/components/common/search-results.vue
```diff
@@ -104,6 +104,9 @@ export default {
         this.searchIsLoading = false
       } else {
         this.searchIsLoading = true
+      },
+      results() {
```
NGPixel, line 108: This code is invalid. You're still within the `search` function.

NGPixel (changes requested):

zemzot:
A foolish mistake my sincerest apologies

zemzot:
✨

PR #5354, opened 2022-06-01
fix: search cleared on enter

zemzot:
When a non-cached query or query with no results is quickly typed, the query is deleted from the search bar.
https://user-images.githubusercontent.com/106627257/171472142-faf5cd7f-3f82-4cc0-9021-56a9c453a7fa.mp4


danieldll099 (approved):

zemzot:
✨

PR #5366, opened 2022-06-04
fix: previous query’s suggestions show up on empty search

zemzot:
### Issue
After deleting a query, suggestions for that query display when the search is cleared.
wiki js-suggestions-for-empty-query

### Solution
Display suggestions only when the query is 2+ characters. 
```template(v-if='search && search.length >= 2 && suggestions && suggestions.length > 0')```

### DIFF
--- a/client/components/common/search-results.vue
+++ b/client/components/common/search-results.vue
@@ -35,7 +35,7 @@
           :length='paginationLength'
           circle
         )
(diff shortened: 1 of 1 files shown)


myml:
Good job! What I missed. https://github.com/requarks/wiki/commit/8be88cb2737cd7e473b1ace4b42189c707f84614

zemzot:
✨

PR #5379, opened 2022-06-08
fix: empty title allowed

zemzot:
## Issues
When creating or editing a page, empty titles can be saved.
## Solution
Created new error code 6015 that will not allow the user to create or save a page with an empty title. Additionally, the save and create button remain stylistically disabled when the title is empty.


zemzot:
✨

PR #5395, opened 2022-06-12
fix: elasticsearch suggestions not displaying

zemzot:
The suggestions for the elasticsearch engine do not display, so the fix now displays elasticsearch suggestions if there are less than 5 hits.


zemzot:
✨

PR #5397, opened 2022-06-12
feat: set search results max hits

zemzot:
## feat: set search results max hits
Added a new field for users to set the maximum number of results returned from the search engine. 
image(7)

### Problem
Currently, if the total hits are larger than the max hits set in algolia, aws, azure, or elasticsearch search engines, when a user attempts to view pages past the max hits, the empty results page is displayed. For example, if the max hit is set to 20, the pages 1-2 will be viewable. However, if the 3rd+ page is clicked on, the empty results page is displayed. 
wiki js-total-hits-display-instead-of-max-hits

### Solution
Display min(results returned size, max hits) instead of total hits.

### DIFF
--- a/server/app/data.yml
+++ b/server/app/data.yml
@@ -96,8 +96,6 @@ defaults:
       maxAge: 600
       methods: 'GET,POST'
       origin: true
-    search:
-      maxHits: 100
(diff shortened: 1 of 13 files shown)


zemzot:
✨

PR #5408, opened 2022-06-15
YDX-10564 [Fix the search to not show email id]

rajatchauhanyti:


NGPixel:
Password fields are meant to be of password type, not text.

PR #5427, opened 2022-06-25
Fix detection of puppeteer

theCalcaholic:
Previously, there seemed to be a copy/paste error: The `isInstalled` check for puppeteer actually checked if the pandoc executable is installed.

Now it checks for the availability of the puppeteer npm module

### DIFF
--- a/server/modules/extensions/puppeteer/ext.js
+++ b/server/modules/extensions/puppeteer/ext.js
@@ -1,4 +1,3 @@
-const cmdExists = require('command-exists')
 const os = require('os')
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


NGPixel:
- The detection is actually a bit more complex. The package will always exist, so just looking if it can be resolved would be incorrect. The package must instead be built specifically for the host platform and the check should detect that it has. This has already been implemented in v3 (see https://github.com/requarks/wiki-vnext/blob/dev/core/modules/extensions/puppeteer/ext.js).

- Extensions are not actually used in v2 and won't be until v3. That UI was a placeholder for what eventually ended up in v3. So there's no point in improving that code right now.

For these reasons, I'll close the PR. Thanks for submitting a PR though!

theCalcaholic:
Ah okay. :)

It would probably be a good idea to improve the wording of the warning in the extensions section a bit. When I read 

> New extensions cannot be installed at the moment. This feature is coming in a future release.

I assumed, that automatic installation of these extensions was not supported, but they would work after manually installing them. I think, something like "Extensions are an upcoming feature that is not supported yet" would be more clear.

PR #5445, opened 2022-07-05
feat(helm): expose external DATABASE_URL

dnplkndll:


NGPixel:
You make other DB_ vars exclusive to having an internal postgres. Some users may want to still use these variables with an external DB, and not the DATABASE_URL format.

+ This would break new deployments using an old values file.

The condition should instead check if a DATABASE_URL value is provided and make the switch only in that scenario.

dnplkndll:
> You make other DB_ vars exclusive to having an internal postgres. Some users may want to still use these variables with an external DB, and not the DATABASE_URL format.
> 
> * This would break new deployments using an old values file.
> 
> The condition should instead check if a DATABASE_URL value is provided and make the switch only in that scenario.

I was wondering about that case. is there any way to spec the host now? or can fullnameOverride be a FQDN?

It is a solution to just pass all and let the app work it out, or is a helper needed?

dev/helm/templates/deployment.yaml
```diff
@@ -53,6 +53,7 @@ spec:
               value: "{{ default "false" .Values.postgresql.ssl }}"
             - name: DB_SSL_CA
               value: "{{ default "" .Values.postgresql.ca }}"
+            {{- if .Values.postgresql.enabled -}}
```
NGPixel, line 56: If DATABASE_URL is defined, all other DB_ vars will be ignored so there's no need for this condition.

dev/helm/values.yaml
```diff
@@ -102,6 +102,9 @@ sideload:
   #  - name: HTTPS_PROXY
   #    value: http://my.proxy.com:3128
 
+externalPostgresql:
```
NGPixel, line 105: Should be commented by default to avoid triggering the DATABASE URL condition.

NGPixel (changes requested):

dev/helm/templates/deployment.yaml
dnplkndll: SSL?
I might be in the wrong place: (not extracted from ?sslmode=require in DATABASE_URL)

https://github.com/Kencove/wiki/blob/main/server/core/db.js#L32

But seems to be tested again. 
https://github.com/Kencove/wiki/blob/main/server/core/db.js#L42

NGPixel: Since dbConfig is a string (from DATABASE_URL), all SSL code afterwards is not actually used for the db connection.

e.g. options are ignored at https://github.com/Kencove/wiki/blob/main/server/core/db.js#L82 because it's not an object.

The string from DATABASE_URL is passed directly at https://github.com/Kencove/wiki/blob/main/server/core/db.js#L133

so any additional params like `?sslmode=require` would be parsed by the DB driver at this point.

NGPixel (approved):

dnplkndll:
@NGPixel  I did not get the [README.md](https://github.com/requarks/wiki/blob/v2.5.285/dev/helm/README.md) update

when using a digital ocean pg service you get the connection string like:
```
postgresql://wikijs:${pass}@${host prefix}.b.db.ondigitalocean.com:25060/wikijs?sslmode=require
```
- needs protocol mapped: postgresql: => postgres: 
- error (SELF_SIGNED_CERT_IN_CHAIN undefined:undefined):  sslOptions.rejectUnauthorized = false 

I see you support DO app builds but looks like uses an internal service versus DBaaS

I was going to test the feature but would need image updates to accept new config:

https://github.com/requarks/wiki/blob/main/dev/build/config.yml

perhaps you have run into this and have a proposal to implement? I easy fix would be map in and case in 
https://github.com/Kencove/wiki/blob/main/server/core/db.js#L59

but a more flexible json string might cover more cases?

NGPixel:
Submit another PR for the changes to the README file and I'll merge it.

For DO, I actually enabled the feature in the past to allow for DBaaS but it didn't work and I didn't investigate further. Looks like you've figured out the changes that would be needed to support it. Again, a PR would be appreciated!

PR #5446, opened 2022-07-06
feat: allow self-signed ssl

dnplkndll:
I could not seem to get around the self signed easily. this works perhaps 2 would be better but I could not get it to work even when passed as a string without the head/footer

https://gist.github.com/zmts/f3306945e90914c4dedda5e0eaf1a279#solution-2

Added
feat: startup probe

at any rate this will work with DO connection string with the change of sslmode=require to ssl=true. could handle but likely a note is sufficient or maybe os obvious.

### DIFF
--- a/dev/helm/templates/deployment.yaml
+++ b/dev/helm/templates/deployment.yaml
@@ -41,9 +41,11 @@ spec:
           env:
             - name: DB_TYPE
               value: postgres
-            {{- if .Values.externalPostgresql.databaseURL }}
+            {{- if (.Values.externalPostgresql).databaseURL }}
             - name: DATABASE_URL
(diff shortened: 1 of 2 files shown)


NGPixel (approved):

dnplkndll:
@NGPixel a secret deployment on the apparent external support using postgresqlHost (wip)

https://gitlab.com/kencove/wiki/-/merge_requests/3/diffs?commit_id=f5c79fd2756b7b8e9221496f2315e2cf2cc20c4d#f8037eadf7a725cbab3a26f103f3de6285d2c0ca

Also I think if your default image is a floating latest you would want an Always pull or the image will not change once on the node. Not planning to PR unless you see a need as our case is covered now with the _URL this was effort in the postgresql.ca approach

PR #5452, opened 2022-07-07
CAS Authentication

SeaLife:
Hey there,

i've implemented the CAS workflow. Only tested CAS3.0 so far but it seems to be working just fine :)

### DIFF
--- a/server/modules/authentication/cas/definition.yml
+++ b/server/modules/authentication/cas/definition.yml
@@ -8,8 +8,22 @@ website: https://apereo.github.io/cas/
 useForm: false
 isAvailable: true
 props:
-  ssoBaseURL: String
-  serverBaseURL: String
-
-  emailAttribute: String
-  displayNameAttribute: String
+  baseUrl:
+    type: String
+    title: Base URL
+    hint: 'Base-URL of your WikiJS (for example: https://wiki.example.com)'
+    order: 1
+  casUrl:
+    type: String
+    title: URL to the CAS Server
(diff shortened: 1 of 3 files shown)


SeaLife:
Requested with https://requarks.canny.io/wiki/p/add-cas-authentication-support

server/modules/authentication/cas/definition.yml
```diff
@@ -6,6 +6,34 @@ logo: https://static.requarks.io/logo/cas.svg
...
+    type: String
+    title: URL to the CAS Server
+    hint: 'Base-URL of the CAS server, including context path. (for example: https://login.company.com/cas)'
+    order: 2
+  casVersion:
+    type: String
+    title: CAS Version
+    hint: 'This provider is only tested with CAS3.0, CAS1.0 is supported by the library but we dont know if that works or not'
+    order: 3
+    default: 'CAS3.0'
```
NGPixel, line 26: Should be a list of values. The user shouldn't have to look up the library to know the possible values.

Also change the hint to just "The version of CAS to use."
If a previous version doesn't work, they can start a new issue or submit a PR for it.

NGPixel (changes requested):

server/modules/authentication/cas/definition.yml
SeaLife: Okay, i added `enum` values for `CAS1.0` and `CAS3.0` and rephrased the hint. :)

NGPixel (approved):

PR #5498, opened 2022-07-21
Update docker-compose file to be usable

waveywaves:
The existing docker-compose file was updated in 2020. This commit
updates this to a usable docker-file which builds wikijs and also
helps use it against a database.


NGPixel:
Closing as your changes don't make any sense. This is for a dev environment, yet you're using a production dockerfile, env settings, force memory limits and point to your own remote context. None of this make any sense.

PR #5499, opened 2022-07-22
feat: add elasticsearch ssl options

Hexaflexagon:
I added ssl options for elasticsearch. This allows the use of elasticsearch with self signed certificates.

image

### DIFF
--- a/server/modules/search/elasticsearch/engine.js
+++ b/server/modules/search/elasticsearch/engine.js
@@ -358,7 +358,7 @@ module.exports = {
 function getTlsOptions(conf) {
   if (!conf.tlsCertPath) {
     return {
-      rejectUnauthorized: conf.verifyTLSCertificate,
+      rejectUnauthorized: conf.verifyTLSCertificate
     }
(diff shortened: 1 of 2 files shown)


server/modules/search/elasticsearch/engine.js
```diff
@@ -351,3 +354,21 @@ module.exports = {
...
+  const caList = []
+  if (conf.verifyTLSCertificate) {
+    caList.push(fs.readFileSync(conf.tlsCertPath))
+  }
+
+  return {
+    rejectUnauthorized: conf.verifyTLSCertificate,
+    ca: caList
+  }
+}
```
NGPixel, line 374: Removing dangling commas

NGPixel (changes requested):

server/modules/search/elasticsearch/engine.js
Hexaflexagon: Done

NGPixel (approved):

PR #5568, opened 2022-08-24
feat: Set groups based on OIDC

fionera:
This is a first try at getting oidc groups to work

### DIFF
--- a/server/modules/authentication/oidc/authentication.js
+++ b/server/modules/authentication/oidc/authentication.js
@@ -29,13 +29,15 @@ module.exports = {
               email: _.get(profile, '_json.' + conf.emailClaim)
             }
           })
-          const groups = _.get(profile, '_json.' + conf.groupsClaim)
-          if (groups) {
-            const groupIDs = Object.values(WIKI.auth.groups)
-              .filter(g => groups.includes(g.name))
-              .map(g => g.id)
-            for (let groupID of groupIDs) {
-              await user.$relatedQuery('groups').relate(groupID)
+          if (conf.mapGroups) {
+            const groups = _.get(profile, '_json.' + conf.groupsClaim)
+            if (groups) {
+              const groupIDs = Object.values(WIKI.auth.groups)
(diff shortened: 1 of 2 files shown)


server/modules/authentication/oidc/authentication.js
```diff
@@ -29,6 +29,15 @@ module.exports = {
               email: _.get(profile, '_json.' + conf.emailClaim)
             }
           })
+          const groups = _.get(profile, '_json.groups')
```
NGPixel, line 32: Make the `groups` path a configurable property.

NGPixel (changes requested):

server/modules/authentication/oidc/authentication.js
fionera: I added a config entry for it :)

fionera:
Ah shoot I just saw that now the Group will be added on every login and they are not deduped. In the WebUI there is the group multiple times now. Also old groups wont be deleted. I will make another PR that removes all groups and then relates to the new ones  or diffs them. Idk yet

fionera:
Could you open an Issue for it? I dont know yet when I have time for it :)

NGPixel:
Fixed in v2.5.2588

cvonderstein:
Hi, I just stumbled upon this and I am wondering, is this implemented now? I am kind of confused by the "Fixed in v2.5.2588" comment, since all I found is v2.5.306. What am I missing here?

voc0der:
I also have this problem still, I don't think it was ever truly resolved.

vlad-patoka:
I implemented a federated Cognito setup, where a Cognito user pool in one AWS account acts as the identity provider (IdP) for a Cognito pool in another account. Wiki.js successfully creates user groups regardless of whether the authentication comes from the local or federated Cognito pool. However, to inject custom claims into the token response, I had to use Cognito extensions - specifically, a Lambda trigger for pre-token generation. Also it required a full refactor of authentication.js to properly handle the integration.

PR #5727, opened 2022-10-10
add authentication module to use a JWT token

lilianabiven:
This module allows a user to authenticate using a jwt token containing a user id, an email and a generation date, for example:
```
{
  "id": "123",
  "email": "johndoe@example.com",
  "iat": 1516239022
}
```
The JWT secret is entered in the user interface. A token expires after 60 minutes, it could be a good idea to also make this duration configurable in the user interface.


lilianabiven:
@NGPixel any interest in this?

slowdream (approved):

PR #5729, opened 2022-10-10
added better AzureAD Support

rubentalstra:
Hello,

Added:
- jobTitle
- location
When a user create an account with AzureAD. 

is there a poseblitiy to add another field to the Config of AzureAD. (if this is okay. i will also update the documentation to work with this.)

Screenshot 2022-10-10 at 19 48 43

### DIFF
--- a/server/models/users.js
+++ b/server/models/users.js
@@ -225,6 +225,18 @@ module.exports = class User extends Model {
       })
     }
 
+    // Parse jobTitle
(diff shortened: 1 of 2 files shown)


rubentalstra:
What do you think?

I have also added the possibility when your "organization" includes a job title and or location you work from. It will be added straight away when you create a new user.

(P.S. I could also update the Azure documentation to work with this.)

PR #5738, opened 2022-10-12
Fix: Overwriting assert with the same name

KainanSu:
Fix bug: As shown in #5599 and #5722, assert will be overwrited when uploading assert with the same name.

A simple way to solve the problem: Add username and unix time before the origin name.


NGPixel:
This is not the desired behavior. Most users don't want to have a username and timestamp added to the filename.

PR #5781, opened 2022-10-24
fix map() call on undefined; fix unrelate() usage

asenchuk:
The changes in v2.5.288 doesn't seem to be tested well. We use Keycloak via OIDC (because we need a group mapping which is absent in Keycloak provider) so I fixed it right in prod (yeah it's sort of things we always love). Sending a PR, hope the fix works for others.

1. `await user.$relatedQuery('groups').select('groups.id')` already returns a groups array, not an object with `groups` property.
2. unrelate() should be used with a where clause


shawnweeks:
I encountered and fixed this same issue today without noticing this PR and I did the exact same thing. +1 for me

NGPixel (approved):

PR #5796, opened 2022-10-31
fix: incompatibility with passport-openidconnect

natsutteatsuiyone:
Fixed this error.

image

https://github.com/jaredhanson/passport-openidconnect/blob/101e0f3739157e6f9f0f027f4c8ee4553d6196ff/lib/strategy.js#L223-L252


natsutteatsuiyone:
See https://github.com/requarks/wiki/pull/5799.

Issue #23, opened 2017-03-04
Remove native compilation dependencies


NGPixel:
Make Wiki.js work without any need for native compilation.

PR #5799, opened 2022-11-01
fix: incompatibility issues with passport-openidconnect@0.1.1

natsutteatsuiyone:
Fixed issues with login and "Map Groups".
image

https://github.com/jaredhanson/passport-openidconnect/blob/101e0f3739157e6f9f0f027f4c8ee4553d6196ff/lib/strategy.js#L210-L237

### DIFF
--- a/server/modules/authentication/oidc/authentication.js
+++ b/server/modules/authentication/oidc/authentication.js
@@ -20,7 +20,9 @@ module.exports = {
         userInfoURL: conf.userInfoURL,
         callbackURL: conf.callbackURL,
         passReqToCallback: true
-      }, async (req, iss, sub, profile, cb) => {
+      }, async (req, iss, uiProfile, idProfile, context, idToken, accessToken, refreshToken, params, cb) => {
+        const profile = Object.assign({}, idProfile, uiProfile);
+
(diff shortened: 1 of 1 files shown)


asenchuk:
I would suggest just removing `sub` param in line #23 - that will be the shortest fix.

natsutteatsuiyone:
> I would suggest just removing `sub` param in line #23 - that will be the shortest fix.

`profile` param never has `_json` property.
So we need `uiProfile` for "Map Groups".

https://github.com/jaredhanson/passport-openidconnect/blob/101e0f3739157e6f9f0f027f4c8ee4553d6196ff/lib/profile.js
https://github.com/jaredhanson/passport-openidconnect/blob/101e0f3739157e6f9f0f027f4c8ee4553d6196ff/lib/strategy.js#L271-L272
https://github.com/jaredhanson/passport-openidconnect/blob/101e0f3739157e6f9f0f027f4c8ee4553d6196ff/lib/strategy.js#L200-L218

asenchuk:
You're right. Have tried your fix already. It works great!

NGPixel (approved):

PR #5808, opened 2022-11-05
Fix login with Keycloak 20 by explicit set OAuth scopes

silicht:
Login failed after updating Keycloak to version 20.0.

Wiki.js showed the "Ooops, somethin went wrong" screen without any error details and Keycloak logged a unauthorized request to the `userinfo` resource.

A default `access_token` requested with an empty scope list may no longer get the OpenID user information. We can enable this by explicitly setting the scope to `openid email profile` to get the required data.

This issue was also described by @aurorasmiles  here: https://github.com/requarks/wiki/discussions/5805

### DIFF
--- a/server/modules/authentication/keycloak/authentication.js
+++ b/server/modules/authentication/keycloak/authentication.js
@@ -20,8 +20,7 @@ module.exports = {
         clientID: conf.clientId,
         clientSecret: conf.clientSecret,
         callbackURL: conf.callbackURL,
-        passReqToCallback: true,
-        scope: 'openid email profile'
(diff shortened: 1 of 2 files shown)


NGPixel:
Put the scopes in the `definition.yml` file instead, e.g.:
https://github.com/requarks/wiki/blob/main/server/modules/authentication/auth0/definition.yml#L10-L13

silicht:
Thanks @NGPixel for the tip, adding the area to the `definition.yml` file seems to fix the problem as well.

NGPixel (approved):

PR #5873, opened 2022-11-19
feat: edit drawio site url

dzruyk:
Add support for drawio in offline mode (according to https://js.wiki/feedback/p/make-drawio-support-for-offline-mode)

New `config.yml` value  `drawioBaseUrl` added. By default it points to https://embed.diagrams.net/, but user can change it to some internal server.

## Howto check
Get all files from https://github.com/jgraph/drawio/tree/dev/src/main/webapp directory
and copy them to some server inside internal network. With docker, you may use `nginx` container.

```
docker run  --publish 8094:80 -v "$PWD/webapp":/usr/share/nginx/html/subpath nginx
```

In that case drawioBaseUrl would look like this
```
drawioBaseUrl: http://127.0.0.1:8094/subpath
```


client/components/editor/editor-modal-drawio.vue
```diff
@@ -2,14 +2,13 @@
   v-card.editor-modal-drawio.animated.fadeIn(flat, tile)
     iframe(
       ref='drawio'
-      src='https://embed.diagrams.net/?embed=1&proto=json&spin=1&saveAndExit=1&noSaveBtn=1&noExitBtn=0'
+      v-bind:src= "drawioBaseUrl()"
```
NGPixel, line 5: - No need to put `v-bind`, just a colon is sufficient.
- Use single quotes

client/components/editor/editor-modal-drawio.vue
```diff
@@ -40,6 +39,12 @@ export default {
     close () {
       this.activeModal = ''
     },
+    drawioBaseUrl() {
+      var params = '?embed=1&proto=json&spin=1&saveAndExit=1&noSaveBtn=1&noExitBtn=0';
+      var url = siteConfig.drawioBaseUrl
+      url += url.endsWith('/') ? '' : '/'
+      return  url + params;
+    },
```
NGPixel, line 47: - Should be a computed property, not a method.
- Don't use `var`, use either const or let
- No semicolons

Could be simplified to:
```js
return `${siteConfig.drawioBaseUrl}?embed=1&proto=json&spin=1&saveAndExit=1&noSaveBtn=1&noExitBtn=0`
```

NGPixel (changes requested):

quadeare:
Hi,

That's a very useful feature for offline instances.

At this moment, we need to build `wikijs` image and `sed` `/wiki/assets/js/editor.js`

```docker
ARG TAG=latest
FROM <private_registry>/requarks/wiki:${TAG}

# Use drawio.private.local instead-of draw.io public website
RUN sed -i 's/embed.diagrams.net/drawio.private.local/g' /wiki/assets/js/editor.js
```

melvin-suter:
@NGPixel What is the status of this issue? It looks like quadeare did the requested changes and the review is pending for over a year.

I got a few customers waiting for this feature

Kvaksrud:
Need this

mvdkleijn:
Hi @NGPixel I realize you're busy, but could I bother you to take a quick look at #5873 since that appears to be ready to merge? Your comments have been resolved and the PR has been waiting on your approval / merge for almost two years now. It'd be very nice to get this in. The PR has no conflicts with `main`.

PR #5921, opened 2022-12-05
Update @mdi/font Version

Ik1497:


NGPixel:
Duplicate of #5233

MDI will not be updated until v3 as it's a breaking change.

PR #5973, opened 2022-12-20
feat: sideloader can import availabilities

topdev-spetermann:
We have a setup with kubernetes where external communication is not allowed. So our solution for localization was to use the offline mode and build a custom image from the base Wiki.js image, where we preload the localization files during the build process of the image. This way works fine, but in the admin ui we can not see the availability of each locale because this information would normally be loaded from the GraphQL endpoint which is of course blocked in our environment.
Our approach currently loads the localization files at the build time of the docker image from the GraphQL endpoint, but the sideloader doesn't support importing the availabilities from the locales. So maybe this extension of the sideloader would be helpful for others who have the same problem.

### DIFF
--- a/server/core/sideloader.js
+++ b/server/core/sideloader.js
@@ -53,7 +53,7 @@ module.exports = {
                   isRTL: locale.isRTL,
                   name: locale.name,
                   nativeName: locale.nativeName,
-                  availability: locale.availability ?? 0
+                  availability: locale.availability || 0
                 }).where('code', locale.code)
               } else {
                 await WIKI.models.locales.query().insert({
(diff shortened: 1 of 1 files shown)


server/core/sideloader.js
```diff
@@ -52,15 +52,17 @@ module.exports = {
                   strings: lcObj,
                   isRTL: locale.isRTL,
                   name: locale.name,
-                  nativeName: locale.nativeName
+                  nativeName: locale.nativeName,
+                  availability: locale.availability ?? 0
```
NGPixel, line 56: Don't use the ?? operator as its not available under Node 12 which has to be supported. Use || instead

NGPixel (changes requested):

server/core/sideloader.js
topdev-spetermann: Fix is pushed

NGPixel (approved):

PR #6014, opened 2023-01-04
git storage: disable color.ui

EricFromCanada:
If Git's `color.ui` has been set to "always" globally, the git storage module will fail to sync because `branches.all` will contain "\u001b[31mremotes/origin/master\u001b[m" instead of "remotes/origin/master", as revealed by `WIKI.logger.info(JSON.stringify(branches.all[0]))`.

### DIFF
--- a/server/modules/storage/git/storage.js
+++ b/server/modules/storage/git/storage.js
@@ -45,9 +45,10 @@ module.exports = {
       await this.git.init()
     }
 
-    // Disable quotePath
+    // Disable quotePath, color output
     // Link https://git-scm.com/docs/git-config#Documentation/git-config.txt-corequotePath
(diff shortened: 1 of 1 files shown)


NGPixel (approved):

Nyxtorm:
Hi,

It seems to me that this PR should have corrected my problem mentioned here #6112 but it is not the case. (if I apply my workaround with `color.ui` instead of `color.branch` it doesn't work, I must absolutely indicate `color.branch = false` as indicated in my discussion in ~/.gitconfig for dedicated WikiJS user.

And if I add in `server/modules/storage/git/storage.js` :

```js
await this.git.raw(['config', '--local', 'color.branch', false])
```
it doesn't seem to be taken into account.

What could cause these parameters to be ignored?

Sorry for my english and thanks for your help!

PR #6031, opened 2023-01-11
Elasticsearch v8.x added

YunusEmreAlps:


YunusEmreAlps:
And Dynamic TOC added


* Screenshot (781)
* Screenshot (782)
* Screenshot (783)

NGPixel:
Please keep PRs to a **single topic**. A dynamic ToC is completely unrelated to the PR original goal.

YunusEmreAlps:
I just wanted to add the newly added features quickly, but I threw the two features through the same PR. Sorry it won't happen again

server/modules/search/elasticsearch/engine.js
```diff
@@ -134,6 +145,7 @@ module.exports = {
           locale: r._source.locale,
           path: r._source.path,
           title: r._source.title,
+          isPublic: r._source.isPublic ? r._source.isPublic : false,
```
NGPixel, line 148: That field doesn't exist... and I'm not sure why it's part of the PR either.

NGPixel (changes requested):

NGPixel:
Looking at the dependencies, 8.x requires Node 14 instead of 12 in previous 7.x and 6.x versions. This could break installations for users still on 12 (which is still supported).

So I'm not going to merge this PR for now. If the node requirements change in the future, I'll re-consider it.

YunusEmreAlps:
Ok, Thank you for your I help. I hope this feature can add soon as possible...

saschabrockel:
That's unlucky that there are compatibility problems. It's the first time I set up Elasticsearch and used version 8 and cannot switch that easily to version 7 now.

> This could break installations for users still on 12 (which is still supported)

Do you mean on local installations?

Maybe the version switch can be done when Wiki.js 3 will be released. Whenever it will be.

YunusEmreAlps:
Screenshot (288)

When I worked with my local environment I didn't realize but I think in the CI/CD pipeline library error deferred this update. Wiki.js main node version and elastic v8 incompatible in CI/CD.

PR #6038, opened 2023-01-12
feat: sqlite3 page content search module

dzruyk:
New search module for sqlite3. (using [fts5](https://www.sqlite.com/fts5.html)).
The module allows  to search by page content like `server/modules/search/postgres` module.

The module allows you to use expressions for searching(similar to `ts-query` [module](https://github.com/requarks/wiki/blob/main/server/modules/search/postgres/engine.js#L1)). Supported types of expressions:

* `a b c d` -- "a" AND "b" AND "c" AND "d"
* `a,b,(c d)` -- "a" OR "b" OR ("c" AND "d")
* `head*`  -- all words with "head" preffix
* a b (!c,!d) -- "a" "b" NOT ("c" AND "d")
* `"foo !bar !baz"`  --  "foo !bar !baz"


dzruyk:
Convert to draft, due some knex troubles with index rebuild
```
Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a transacting(trx) call?
```

dzruyk:
The `rebuild` function has been rewritten.

It looks like knex only supports one sqlite database connection. I believe the error was caused by multiple knex function calls inside the pipeline.

```
await pipeline(
  WIKI.models.knex.column(...).select(),
  
  new stream.Transform({
    objectMode: true,
    transform: async (page, enc, cb) => {
      await WIKI.models.knex.raw(...).
      cb()
    }
  })
)
```

NGPixel:
I cleaned up the engine.js file, but the match-query.js needs some reformatting too, so that it conforms to [StandardJS](https://standardjs.com/) (e.g. always use braces for if statements, use const when variable is not reassigned afterwards, no dangling commas, etc.)

dzruyk:
Thx, I cleaned up the `match-query.js`.

Almost all warnings have been fixed.

There are still some warnings at the definition of [this.knownLexemes](https://github.com/requarks/wiki/blob/c47e63383e3ec5b026ebde370d09de5fb240fb03/server/modules/search/sqlite3/match-query.js#L124).
But I doubt whether it is worth rewriting it and how critical these warnings are.

```
  match-query.js:125:19: Object properties must go on a new line if they aren't all on the same line. (object-property-newline)
  match-query.js:125:31: Object properties must go on a new line if they aren't all on the same line. (object-property-newline)
  match-query.js:125:31: Unnecessarily quoted property 'not' found. (quote-props)
  match-query.js:126:19: Object properties must go on a new line if they aren't all on the same line. (object-property-newline)
  match-query.js:126:19: Unnecessarily quoted property 'and' found. (quote-props)
  match-query.js:127:18: Object properties must go on a new line if they aren't all on the same line. (object-property-newline)
  match-query.js:127:18: Unnecessarily quoted property 'or' found. (quote-props)
  match-query.js:127:30: Object properties must go on a new line if they aren't all on the same line. (object-property-newline)
```

PR #6052, opened 2023-01-15
Dynamic ToC added.

YunusEmreAlps:
Screenshot (786)
Screenshot (787)


NGPixel:
There're 2 major issues that need to be addressed:

- Calling a method on the scroll event, without any debounce, is highly inefficient and can lag on complex pages for users with under-powered PCs.
- You're modifying the DOM directly, which is a big no. You should add the classes dynamically using Vue, not by modifying the DOM.

YunusEmreAlps:
Hi @NGPixel, can you review last changes?

PR #6053, opened 2023-01-15
OAuth2: add groups mapping

utix:
Allow to map groups.
This a copy paste of the feature of oidc authentication


gueldenstone:
Would be great if this would get merged!

NGPixel (approved):

PR #6055, opened 2023-01-16
Id instead of path for pages

nafisgi:
I suggest using the id in the router instead of the path, because at the moment there are problems using the path (for example, when moving a page, links to this page do not change) and this solution will be resolve a lot of problems in the future


NGPixel:
This is introducing a whole other set of problems and force users to link via IDs which is not what most users want.

You can already reference pages by IDs directly using `/i/123` where 123 is the page ID.

PR #6057, opened 2023-01-16
fix: Helm Chart - Sideloading

B3nBeng:
Context : Wiki container is isolated from the internet, so i enabled sideloading.

Problems : 

1. The init container don't have volumeMounts, so wiki container can't get the locales files previously downloaded.
2. Wiki container don't use locales files in "sideload" folder because we can't set the option "offline" to true.

Fixs : 

1. Added volumeMounts for the init container, for mount an "emptyDir" with the wiki container, for example.
2. 
- Added var "OFFLINE_ACTIVE" for set offline option.
- Added offline var in values.yaml

**NB :** Implies to add "offline: $(OFFLINE_ACTIVE)" into /wiki/config.yml in the docker image.

### DIFF
--- a/dev/helm/templates/deployment.yaml
+++ b/dev/helm/templates/deployment.yaml
@@ -32,6 +32,10 @@ spec:
           command: [ "sh", "-c" ]
           args: [ "mkdir -p /wiki/data/sideload && git clone --depth=1 {{ .Values.sideload.repoURL }} /wiki/data/sideload/" ]
       {{- end }}
+    {{- with .Values.volumeMounts }}
+          volumeMounts:
+            {{- toYaml . | nindent 12 }}
+      {{- end }}
       containers:
         - name: {{ .Chart.Name }}
           securityContext:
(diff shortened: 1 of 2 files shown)


NGPixel:
> NB : Implies to add "offline: $(OFFLINE_ACTIVE)" into /wiki/config.yml in the docker image.

Then it should be part of the PR in https://github.com/requarks/wiki/blob/main/dev/build/config.yml
and tested to make sure it still works if no env var is present.

B3nBeng:
> > NB : Implies to add "offline: $(OFFLINE_ACTIVE)" into /wiki/config.yml in the docker image.
> 
> Then it should be part of the PR in https://github.com/requarks/wiki/blob/main/dev/build/config.yml and tested to make sure it still works if no env var is present.

It's done. I added a default value "false". It's works correctly.

NGPixel:
What about non-kubernetes? That config is used for docker images. It needs to work with no value as well.

B3nBeng:
> What about non-kubernetes? That config is used for docker images. It needs to work with no value as well.

Still works in Docker with no values for offline. 
Tested with this docker-compose : https://docs.requarks.io/install/docker#using-docker-compose

bdsoha:
@NGPixel Any updates?

PR #6072, opened 2023-01-20
Elasticsearch APM RUM SSL cert added.

YunusEmreAlps:


NGPixel:
I don't think that makes sense? This runs on the client, not the server.

Can you point to the documentation where this parameter is accepted?

YunusEmreAlps:
https://www.elastic.co/guide/en/apm/guide/master/ssl-setup.html

NGPixel:
That's for an APM server, not the APM RUM js client.

PR #6188, opened 2023-02-21
feat: use title as short url

myml:
The wiki should hava short url.

Pages can be view using keyword instead of long file path.

like https://zh.wikipedia.org/wiki/JavaScript instead of https://zh.wikipedia.org/wiki/Programming_language/Scripting_languages/JavaScript.

### DIFF
--- a/server/controllers/common.js
+++ b/server/controllers/common.js
@@ -562,6 +562,10 @@ router.get('/*', async (req, res, next) => {
         _.set(res.locals, 'pageMeta.title', 'Welcome')
         res.render('welcome', { locale: pageArgs.locale })
       } else {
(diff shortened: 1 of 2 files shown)


NGPixel:
This would lead to too many unexpected behaviors (e.g. multiple pages with the same title, conflicts with path being identical with a page title, revealing private page paths using a page title, etc.) and would confuse most users.

You're not forced to use long or nested paths... You can create your all your pages at root level and achieve the same result.

v3 will have proper support for redirects which should address your use case.

myml:
@NGPixel Is there a roadmap for v3?

NGPixel:
See https://blog.js.wiki/

PR #6204, opened 2023-02-24
feat(helm): allow usage of additional env vars / configmap / secrets

MickaelCa:
This PR makes it possible to load external environmenent variables.
On my cluster, I'm using [Postgres Operator](https://access.crunchydata.com/documentation/postgres-operator/latest/) by CrunchyData, I need to be able to load the generated database secrets directly rather than copying it in the values.

### DIFF
--- a/dev/helm/templates/deployment.yaml
+++ b/dev/helm/templates/deployment.yaml
@@ -39,6 +39,9 @@ spec:
           image: "{{ .Values.image.repository }}:{{ default "latest" .Values.image.tag }}"
           imagePullPolicy: {{ default "IfNotPresent" .Values.image.imagePullPolicy }}
           env:
+            {{- with .Values.extraEnv.env }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
             - name: DB_TYPE
               value: postgres
             {{- if (.Values.externalPostgresql).databaseURL }}
@@ -46,7 +49,8 @@ spec:
               value: {{ .Values.externalPostgresql.databaseURL }}
             - name: NODE_TLS_REJECT_UNAUTHORIZED
               value: {{ default "1" .Values.externalPostgresql.NODE_TLS_REJECT_UNAUTHORIZED | quote }}
-            {{- else }}
+            {{- end }}
+            {{- if .Values.postgresql.enabled}}
             - name: DB_HOST
               value: {{ template "wiki.postgresql.host" . }}
             - name: DB_PORT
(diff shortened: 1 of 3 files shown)


dev/helm/values.yaml
```diff
@@ -98,6 +98,22 @@ tolerations: []
...
+#      value: "sample value"
+#    - name: SAMPLE_ENV_FROM_SECRET
+#      valueFrom:
+#        secretKeyRef:
+#          name: 'secret-name'
+#          key: 'secret-key'
+  secrets: []
+#    - secret-name
+#    - another-secret-name
+  configmaps: []
```
bdsoha, line 113: IMO, this should be camelCase, as it is found in the k8s API *(`configMapRef`)*.

```suggestion
  configMaps: []
```

dev/helm/templates/deployment.yaml
```diff
@@ -71,6 +75,19 @@ spec:
             {{- end }}
             - name: HA_ACTIVE
               value: {{ .Values.replicaCount | int | le 2 | quote }}
+          envFrom:
+            {{- with .Values.extraEnv.configmaps }}
```
bdsoha, line 79:
```suggestion
            {{- with .Values.extraEnv.configMaps }}
```

bdsoha (changes requested):
IMO, this should be camelCase, as it is found in the k8s API *(`configMapRef`)*.

dev/helm/templates/deployment.yaml
```diff
@@ -39,14 +39,18 @@ spec:
...
             - name: DB_TYPE
               value: postgres
             {{- if (.Values.externalPostgresql).databaseURL }}
             - name: DATABASE_URL
               value: {{ .Values.externalPostgresql.databaseURL }}
             - name: NODE_TLS_REJECT_UNAUTHORIZED
               value: {{ default "1" .Values.externalPostgresql.NODE_TLS_REJECT_UNAUTHORIZED | quote }}
-            {{- else }}
+            {{- end }}
+            {{- if .Values.postgresql.enabled}}
```
stavros-k, line 53:
```suggestion
            {{- if .Values.postgresql.enabled }}
```

dev/helm/templates/deployment.yaml
```diff
@@ -39,14 +39,18 @@ spec:
           image: "{{ .Values.image.repository }}:{{ default "latest" .Values.image.tag }}"
           imagePullPolicy: {{ default "IfNotPresent" .Values.image.imagePullPolicy }}
           env:
+            {{- with .Values.extraEnv.env }}
```
stavros-k, line 42: I'd suggest checking if `extraEnv` is not null first.
On upgrade, it won't exist in values.yaml and will most likely error with `{}.env` `nil` pointer.

PR #6217, opened 2023-03-01
Allows setting the NODE_EXTRA_CA_CERTS variable

dostalradim:
We have the keycloak instance signed with our internal CA and I do not like disabling of certs trusting. So we just need to append extra CA certs.


NGPixel:
I would prefer the option be named `nodeExtraCaCerts` rather than `NODE_EXTRA_CA_CERTS` to keep things consistent.

dostalradim:
That makes sense. I was inspired by the variable NODE_TLS_REJECT_UNAUTHORIZED from postgresql chart :). I am gonna change it.

NGPixel (approved):

PR #6238, opened 2023-03-08
Update auth strategies with new properties

icsinfo:
Hopefully this pr will address [#6139](https://github.com/requarks/wiki/discussions/6139).

This will detect any property additions to an auth strategy that are added by an update, and write the new defaults into the database. It does this by using the existing method to append the new strategy properties, then write the strategy that method loaded back to the database with the new defaults.

My only fear about this solution is it could break someone's install if a new auth strategy property has a bad default that causes an in use strategy to not work. This will not update strategy properties that have new defaults set, only in use strategies that have new properties added by an update.

Ignore the launch.json, I wasn't trying to put that in there 😅.

### DIFF
--- a/dev/helm/values.yaml
+++ b/dev/helm/values.yaml
@@ -102,7 +102,7 @@ volumeMounts: []
 
 volumes: []
 
-# This will allow us to install locales even without internet access using a initContainer & wikjs "sideloading"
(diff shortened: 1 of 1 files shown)


NGPixel:
Discard the `.vscode/launch.json` changes in your PR.

PR #6266, opened 2023-03-16
Elasticsearch hits# fix

IBRMHIRI:
Elasticsearch may provide wrong total hits number, which may cause in the frontend generating too many paginations and results with no references.
Counting the length of the returned results from Elasticsearch may solve the problem.

### DIFF
--- a/server/modules/search/elasticsearch/engine.js
+++ b/server/modules/search/elasticsearch/engine.js
@@ -137,7 +137,7 @@ module.exports = {
           description: r._source.description
         })),
         suggestions: _.reject(_.get(results, 'suggest.suggestions', []).map(s => _.get(s, 'options[0].text', false)), s => !s),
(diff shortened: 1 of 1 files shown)


server/modules/search/elasticsearch/engine.js
```diff
@@ -137,7 +137,7 @@ module.exports = {
           description: r._source.description
         })),
         suggestions: _.reject(_.get(results, 'suggest.suggestions', []).map(s => _.get(s, 'options[0].text', false)), s => !s),
-        totalHits: _.get(results, 'body.hits.total.value', _.get(results, 'body.hits.total', 0))
+        totalHits: _.get(results, 'body.hits.hits', []).map(r => r.length)
```
NGPixel, line 140: The `.map()` doesn't make sense. Your returning an array of lengths, instead of just the length...

```
totalHits: _.get(results, 'body.hits.hits', []).length
```

NGPixel (changes requested):

PR #6273, opened 2023-03-17
[Fix] List paragraphs inline 100%, image into lists

pgi-jsanchez:
Hello. 

Editing in a private wiki, I try to use images to be part of a list, the issue is that those got aligned with the text, so I tested with the css to fix it. I also thought to change it with `display:block;`, but if there's a reason to use inline-block for this, this change still will respect that and fix this issue. I include this tests images.

Before:
Captura de pantalla-20230317151347-784x340

After:
Captura de pantalla-20230317151307-908x422

### DIFF
--- a/client/themes/default/scss/app.scss
+++ b/client/themes/default/scss/app.scss
@@ -685,6 +685,7 @@
         display:inline-block;
         vertical-align:top;
         padding-top:0;
+	width: 100%;
       }
     }
   }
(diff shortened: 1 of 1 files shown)


NGPixel:
This makes many assumptions that are incorrect.

- Your change applies to all element types, regardless of whether they are an image or not.
- What if the image is smaller?

pgi-jsanchez:
Does not apply to all, only for lists first paragraph, that actually contains the text. Adding a css that changes this paragraph when next paragraph contains an image, might fail if there are nested lists or other text, so with this, can help in this situation.

I must admit that images changes the sizes a little bit if using =%x, but I tested this change and using this parameter, and works just fine.

Captura de pantalla-20230317171817-745x509

NGPixel (approved):

PR #6299, opened 2023-03-27
Sync groups with SAML provider

aelgasser:
This PR enables syncing groups from the IdP in SAML strategies.

It adds two new fields to the SAML configuration: 

image

The field's value is sent by the IdP into `profile.attributes` through the `Group attributes` field when enabled (see below my setup with Jumpcloud, all SAML-compliant providers such as Okta or Keycloak expose such attribute): 

image

Once connected, the groups coming from the IdP will overwrite the groups of the user in the Wiki (I copied the behavior from the LDAP group sync; which is a good assumption as the enterprise directory must be the ground truth for user access in that case).

One final note, to allow JIT provisioning and IdP-initiated login, wiki admins must enable the "Allow self-registration" option in the SAML strategy and leave the `Assign to group` field blank (it'll be overwritten anyway): 

image


aelgasser:
@NGPixel do you think you can review and approve this?

enjoeiluis (approved):

elomarcorreia (approved):

andreleoni (approved):

aelgasser:
Hi @NGPixel , @andreleoni , @elomarcorreia , @enjoeiluis, is there anything missing to merge this PR?

NGPixel (approved):

NGPixel:
Thanks! Sorry for the delay.

aelgasser:
Hi @NGPixel thanks for merging!

I just worked on a modification of this feature (here https://github.com/aelgasser/wiki-js/pull/1/files) to control how the group assignments are synchronized. In my case, I wanted to keep the default admin group of WikiJS assigned when a user signs on, and I couldn't use the "Administrators" group name in our identity provider as it's already used to define the administrators of our organization and I needed to separate the WikiJs admins from the organization admins.

Is it worth adding to the main codebase (and probably to be added to the LDAP implementation, though I don't have any way to test it locally) or should I just keep it for me?

PR #6313, opened 2023-04-02
fix: Fix the uploadInfo text from static text to dynamic

ckfear:
The previous text is static, if the configuration is modified, the display will be wrong.
`.caption.grey--text.text-darken-2 Max 10 files, 5 MB each`
Fix the uploadInfo text from static text to dynamic.


NGPixel:
That won't work for non-admin users and will throw an error.

ckfear:
I added an api to provide access to uploaded configuration information, and added non-administrator permissions.

PR #6412, opened 2023-05-05
feat: update analytics module umami

CDN18:
Umami has entered the next major version, introducing a breaking change to its script structure.

See: https://umami.is/docs/migrate-v1-v2

### DIFF
--- a/server/modules/analytics/umami/code.yml
+++ b/server/modules/analytics/umami/code.yml
@@ -1,2 +1,2 @@
 head: |
-  <script async defer data-website-id="{{websiteID}}" src="{{url}}/umami.js"></script>
+  <script async src="{{url}}/script.js" data-website-id="{{websiteID}}"></script>


NGPixel:
What if someone upgrades to the new Wiki.js version but still use v1? Will their analytics suddenly stop working?

CDN18:
Yes, v2 and v1 are two incompatible versions.

If we want to allow users to choose between v1 and v2 themselves, we can either
- create another analysis module 
- add a setting field that allows users to enter the `script name` (`umami.js` for v1 and `script.js` for v2)

pxlfrk:
+1 for this update - using umami in v2 & can't track my wiki page views :)

NGPixel:
A dedicated module for v2 would be preferred. There's currently no conditional capabilities for analytic modules.

CDN18:
Okay, I'll close this and create another one.

PR #6514, opened 2023-06-19
Unique asset folder

Tak-MK:


Tak-MK:
My fault; was trying to do a PR in my own repo :P

PR #6587, opened 2023-07-20
html-image-prefetch now works with binary images

fjtc:
This patch allows PNG files to work when `html-image-prefetch` is enabled. More details about this issue can be found in  https://github.com/requarks/wiki/discussions/6586.


Divy0007 (approved):
Looks Good

Blessium:
LGTM!!!

PR #6633, opened 2023-08-06
Add docker compose example to run with YugabyteDB

FranckPachot:
This adds a Docker Compose example to run with [YugabyteDB](https://www.yugabyte.com/) (Open Source PostgreSQL compatible Distributed SQL database) instead of PostgreSQL. Deployments on YugabyteDB are compatible with PostgreSQL, and add cloud-native resilience with horizontal scalability


NGPixel:
This would be better suited as a guide in https://github.com/requarks/wiki/discussions/categories/guides

PR #6708, opened 2023-09-07
Fix postgres db custom schema option not working

bergerkiller:
We encountered an issue when trying to configure a postgres DB schema, which was needed for integration with supabase. This was caused by knex migration scripts not having this same schema configured, and this custom schema not being created at startup.

This PR fixes both issues and allows for a successful startup and initialization of the wiki server.

I did notice that this option, while it exists in the code, isn't documented in the docs. It is set by using a config.yml with db.schema set, and I don't think it has an env parsing default. This might be something worth improving in the future.

### DIFF
--- a/server/core/db.js
+++ b/server/core/db.js
@@ -190,7 +190,15 @@ module.exports = {
       // -> Create DB Schema if different than the default 'public'
       async createDefaultSchema () {
         if (WIKI.config.db.schema && WIKI.config.db.schema !== 'public') {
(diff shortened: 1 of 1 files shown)


bergerkiller:
Added a change to only create the custom schema if it doesn't already exist. CREATE IF NOT EXISTS does not work by itself because the user requires CREATE permission on the database for it to work. When granting limited priviliges to the wiki database user this is undesirable.

PR #6775, opened 2023-10-12
Update Azure Application Insights JS to v3

nwcm:
Updates the analytics engine for Azure Application Insights to their v3

https://github.com/Microsoft/ApplicationInsights-JS


NGPixel:
Is v2 being deprecated? If not, v3 should be a different module to avoid breaking changes in existing installations.

PR #6778, opened 2023-10-13
Support generating sitemap for SEO

lawrenceching:
WIP.
Good to have a early review.

### DIFF
--- a/config.sample.yml
+++ b/config.sample.yml
@@ -145,3 +145,9 @@ dataPath: ./data
 # file uploads.
 
 bodyParserLimit: 5mb
+
+
+seo:
+  sitemap:
+    enabled: false # sitemap is experimental feature, do not enable in production
+    cacheExpireTime: 3600 # in seconds
--- a/package.json
+++ b/package.json
@@ -176,6 +176,7 @@
     "semver": "7.3.8",
     "serve-favicon": "2.5.0",
     "simple-git": "3.16.0",
+    "sitemap": "7.1.1",
     "solr-node": "1.2.1",
     "sqlite3": "5.1.4",
     "ssh2": "1.11.0",
--- a/server/controllers/common.js
+++ b/server/controllers/common.js
@@ -5,6 +5,7 @@ const _ = require('lodash')
 const CleanCSS = require('clean-css')
 const moment = require('moment')
 const qs = require('querystring')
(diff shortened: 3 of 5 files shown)


BPowell76:
Any chance on getting this merged? I'd like to add a sitemap to my wiki as I'm using it for my public-facing website. For the time being I can generate a sitemap by hand, but the more pages I have that's going to be tedious and annoying.

BPowell76:
@lawrenceching Some things we should add to keep the sitemap compliant, is have an option to set the priority page in the sitemap. Valid values are 0.0 to 1.0. Another thing to add is a change frequency, but a default value could be assigned for monthly or weekly.

Adding the `<priority>` component to the sitemap and page properties editor will require a change to the database schema by adding a new column.

lawrenceching:
@BPowell76 
I used a default priority and change frequency for every page.
And allow users to change them in config file.

@NGPixel 
I think my change is ready for review. Could you have a look, especially at the permission checking part?

BPowell76:
I don't know if this is going to be merged, but I figured out how to generate a `sitemap.txt` file in the Linux terminal doing the following and using cron jobs to sync everything:

```bash
# generate-sitemap.sh
#
# database name for me is "wiki", replace accordingly

# query database and output results to file
psql -U postgres -d wiki < /var/lib/postgresql/select-published-pages.sql > sitemap.txt

# count rows and delete the first and last 2 rows
# the text output has a column name and a ascii table row line "-------"
# the last two rows for me were the number of rows returned and a blank line
ROWCOUNT=$(sed -n "$=" sitemap.txt)
SECLASTROW=$(($ROWCOUNT-1))

sed -i "${SECLASTROW},${ROWCOUNT}d" sitemap.txt
sed -i "1,2d" sitemap.txt
```

```sql
-- select-published-pages.sql
-- 
-- annoyingly, psql doesn't like camel case column names and prefers snake case

SELECT CONCAT('https://[insert_your_url_here]',path) AS "URL"
FROM pages
WHERE "isPublished"='t'
ORDER BY path ASC;
```

joe-speedboat:
Hi, i just setup my wikijs with this as cronjob on k3s.
does a nice job and is some kind of nice and NOT intrusive to original setup:

https://github.com/joe-speedboat/kube.wikijs

PR #6784, opened 2023-10-16
fix: missing css style 'display: block;' for markdown code block in comments module

hotime:
Add css style 'display: block;' for markdown code block in comments module, to ensure that the code block can be displayed properly.


hotime:
Without the css style of "display: block;", the markdown code block in the comments module will appear as follows:

PR #6813, opened 2023-11-04
Fix items to be of type array

MangaOtaku:
JSON stringifying this causes an error due to expecting type of array.


NGPixel:
Thanks!

PR #6904, opened 2023-12-25
feat: Add Elasticsearch 8.x support

jacobacon:
Modify ElasticSearch search module to support v8.x.

Changes:
- Added 8.x to enum in definition.yml
- Add ElasticSearch 8.x client to dependencies
- Remove 6.x specific _doc mapping definitions in engine.js
- Add 8.x specific mappings
- Modify logic to handle differences between 6.x 7.x / 8.x client responses.

6.x is EOL and no longer supported by ElasticSearch; Some of the logic can be further simplified if support for 6.x is dropped.

Tests:
I've tested fairly extensively with 8.x creating, deleting and renaming pages to validate indices are created. I have not done any regression testing against 6.x or 7.x ElasticSearch.


server/modules/search/elasticsearch/engine.js
```diff
@@ -182,7 +219,6 @@ module.exports = {
   async created(page) {
     await this.client.index({
       index: this.config.indexName,
-      type: '_doc',
```
NGPixel, line 185: Does removing the `type` field in all mutation methods affect 6.x and 7.x?

jacobacon: The docs indicate type was changed with 6.x and deprecated with 7.x to prepare for removal in 8.x. I had assumed it wasn't necessary with 6.x, but seems like it is still required in some situations.

Looks like more special cases are needed to continue 6.x support and add 8.x.

skjoedt:
What is missing for this to land in main?

jacobacon:
> What is missing for this to land in main?

Code wise should be complete now. I've tested with ES 7.x, and 8.x and all features seem to be working as expected. I don't have a 6.x cluster setup, so I wasn't able to test. 

@NGPixel, I modified the code to keep the type field for 6.x and 7.x, and merged in upstream changes. Can you review?

xlammertink:
Hi all, it seems that Wiki.JS is not supporting ElasticSearch version 8 yet, while this version has been around since 2021. Looking at this PR it seems that everything is in place to introduce the support of ElasticSearch 8. @NGPixel Could you review this PR?

NGPixel (approved):

PR #6946, opened 2024-01-23
feat: option to use sqlite

ckt114:


NGPixel:
Thanks but using SQLite in a high-availability environment makes no sense. You'll run into caching issues and race conditions as there's no pub/sub between instances. Only PostgreSQL has this capability which is why it's the requirement for Kubernetes deployments.

ckt114:
I’m running a small home lab wiki so I don’t need PostgreSQL. What’s the harm to have this option in the chart? Not everyone runs in big environments.

NGPixel:
The chart is for Kubernetes which implies high availability capability. By offering sqlite, there's the possible confusion that you can run multiple replicas which is not true. I'd rather not add a misleading option to the chart.

If you want to run with sqlite, then use the docker image directly or create your own basic chart.

PR #7004, opened 2024-02-19
Fix for server rendering

LarsK1:


NGPixel:
This PR makes absolutely no sense.

PR #7165, opened 2024-05-11
containerPort can be specified in helm chart

vampire-yuta:
## Changed
Hi 😄 

In helmchart, containerPort can be specified of deployments.


NGPixel:
The port is always 3000 when using the docker image and I don't see how making it customizable is useful? This is an internal port and doesn't reflect what the service port is.

Closing as this change doesn't make sense to me.

PR #7178, opened 2024-05-17
Added option for embedding plantuml diagram as an object tag

oneacik:
Follow up of: https://github.com/requarks/wiki/pull/2754 and https://github.com/requarks/wiki/discussions/3597

Summary:
Added option to allow plantuml diagrams in svg format to be rendered as <object> tags.

This PR does the following:
- Adds renderSvgAsObject option
- Adds branches in editor and server renderer

Tests:
✅ Links in plantuml work in rendered page when option is enabled, also with enabled Security Plugin.
❌ Links in plantuml don't work in editing mode - seem to be sanitised?. Also configuration options are not available in editor renderer.

Tested Using simple markdown:
``````
# Header

```plantuml
:[[http://example.com TEST]];

```
``````


oneacik:
Closing for now as deploying on production leads to differen outcomes...

PR #7188, opened 2024-05-24
add healthcheck for posgres docker

bgogri:
This is a PR to add healthchek to posgres in docker compose


NGPixel:
Thanks but this is not really useful for a deployment where postgres doesn't have high-availability anyway? This would be better suited in a kubernetes deployment.

In this case, in makes the example docker-compose more complex with little to no actual benefits. The example is meant to be as simple as possible.

PR #7205, opened 2024-06-03
converting HTML to Markdown - Emojis

markus-96:
Hi,

When I have a page in HTML with embedded emojis and I convert them into Markdown, the emojis are always embedded as a picture, filling in the whole width of the page, what breaks the entire page.

For example:

HTML:
```html
 <p><img src="/_assets/svg/twemoji/1f34e.svg" alt="🍎" draggable="false" class="emoji"></p>
```
gets converted to:
```md
![🍎](/_assets/svg/twemoji/1f34e.svg)
```
instead of:
```md
:apple:
```

The map in server/helpers/emoji-data/emoji-reverse-lookup.js was generated in python using this script:

```py
emoji: dict[str: str] = {
    "100": "💯",
    "1234": "🔢",
    "grinning": "😀",
    [...]
}

result: dict[str, str] = {}

for key, value in emoji.items():
    ret_arr: list[str] = []
    first: bool = True
    for c in value:
        code_point_hex = hex(ord(c))
        code_point_hex_str = code_point_hex[2:]
        ret_arr.append(code_point_hex_str)
    result.update({"-".join(ret_arr): key})

for key, value in result.items():
    print(f"'{key}': '{value}',")
```

If you find a way to use the map in markdown-it-emoji/lib/data/full.js directly at runtime, please tell me. This could be much more reliable.


NGPixel:
or you could simply disable the Twemoji rendering step under Admin > Rendering...?

markus-96:
Please, consider reviewing my PR. Disabling emojis is a solution, but not a very good one, if the users actually want to use emoji tags.

NGPixel:
This has nothing to do with disabling emojis. The Twemoji stage replaces emojis with images, which your PR attempts to reverse. Just disable that stage in the first place instead...

markus-96:
as far as I can tell, twemoji is always applied if you use emoji tags, for example `:apple:`, and Emoji in Admin->Rendering->markdown is activated (deactivating it results in no render at all of the emoji tag, as expected). Even disabling Twemoji in Admin->Rendering->html does not change any behaviour, the twemoji svg is always served. I did not find any other Emoji related setting.

PR #7273, opened 2024-07-23
fix: duplicate key value violates unique constraint "userGroups_pkey"

YetheSamartaka:
This fixes the 
`GraphQL error: insert into "userGroups" ("groupId", "userId") values ($1, $2) returning "groupId" - duplicate key value violates unique constraint "userGroups_pkey" `
when userGroups table was modified outside of web interface and there is already existing entry with ID that is the same of what is about to be added when assigning group to user.

It checks for the unique ID and increments it based on existing maxId.

### DIFF
--- a/server/graph/resolvers/group.js
+++ b/server/graph/resolvers/group.js
@@ -71,8 +71,20 @@ module.exports = {
         throw new gql.GraphQLError('User is already assigned to group.')
       }
 
-      // Assign user to group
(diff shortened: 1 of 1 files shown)


NGPixel:
The issue here is really your external script inserting rows with manually defined IDs instead of letting the database provide them. You're proposing a fix for a problem you created externally.

Wiki.js is using the table built-in autoincrement id. If you're inserting rows by explicitly defining the row ID, you're doing it wrong.

Issue #1292, opened 2019-12-07
Two letter virtual folders mistaken for locales 


invisiblethreat:
When linking to something before it exists, like `db/postgres` from "home" and then clicking through to the dialogue to create the new page, you end up with an unlocalized page creation dialog that you can still use. Upon clicking through and creating the page, when you attempt to save, you receive an SQL error. This is potentially due to the `db` section being mistaken for the locale.

**To Reproduce**
1. Go to 'home'
2. Edit the page to add a link to a new "short"  virtual folder/page that does not yet exist, like `db/postgres`
3. Click through the creation dialog
4. Attempt to save the page and capture the error

**Expected behavior**
Throw an error for unsupported locale/action. The modal for creating a new pages does not permit short paths that would conflict with this locales, which may be good enough, but I wanted to raise this issue if folks though it would be helpful to trap it.

**Screenshots**
![image](https://user-images.githubusercontent.com/2525006/70379154-487b7900-18ff-11ea-9288-2f2915490faa.png)

![image](https://user-images.githubusercontent.com/2525006/70379162-5df0a300-18ff-11ea-928c-55f5418dc554.png)

![image](https://user-images.githubusercontent.com/2525006/70379168-6c3ebf00-18ff-11ea-9036-49230e0d1e18.png)

![image](https://user-images.githubusercontent.com/2525006/70379176-94c6b900-18ff-11ea-8d58-479fb3a4f63d.png)

**Host Info (please complete the following information):**
 - OS: Docker, requarks/wiki:2
 - Wiki.js version: 2.0.12
 - Database engine: MariaDB 10.4.7-MariaDB-1:10.4.7+maria~bionic

**Additional context**
Again, this is an edge case that I'm not sure that many folks will hit, so it might not be worth the code to fix/prevent, but I wanted to let you know. Thanks for the cool project!

mielipuolinen:
Can confirm this issue too, run into it few times before I realized what was happening. :)

Pages can be renamed to two character path names which causes issues too and things start to act weird.

juliendms:
There is also an issue with Assets folder with only two letters. You can add pictures in those folders, but when you add them on the page, it returns a 404 error instead of loading the images.

BPowell76:
[...]

PR #7440, opened 2024-11-17
Bug: Fix issue https://github.com/requarks/wiki/issues/1292

BPowell76:
Fixed issue https://github.com/requarks/wiki/issues/1292 with new page creation being incorrectly caught in Regex pattern matching that didn't match a language locale.

The server page helper regex pattern now excludes everything that match a living language as recognized by ISO 639-1. New pages can now be created.

In resolving this issue, the locale Regex pattern matching was removed from the new page modal as it adds the locale to the root of the page path. This made the regex matching redundant.

image

image

image


jarrodlovas:
@BPowell76 Did you close this PR  since it's been resolved in a patch or is the issue still open for someone to resolve?

PR #7445, opened 2024-11-19
Improved OIDC authentication

Zweisteinium:
## Improve OIDC Authentication

**Summary:**
- **Simplified Configuration:** Now only requires `wellKnownURL` to auto-fetch OIDC endpoints.
- **Claim Merging:** Added `mergeIdTokenClaims` option to merge claims from the ID token with the user profile.

**Why:**
- **Easier Setup:** Reduces configuration complexity by leveraging OIDC discovery, also lowering the risk of misconfiguration
- **Accurate Group Mapping:** I actually had massive issues with the group mapping, because with Azure Entra ID the 'roles' claim used to assign roles/groups is inside the ID token. But since the ID token is ignored it was impossible to get this working.

One dependency was added (jwks-rsa) since it is mandatory to verify the ID token

Hope you find some time to look into the PR give some feedback! :)

Screenshot_20241119_183223


NGPixel:
What if the oidc server doesn't have a .well-known endpoint? It should still allows to specify the endpoints manually.

H3xaChad:
I actually wasn't fully aware that the discovery endpoint is optional.
For representing this in the settings, my thought would be to add the well-known URL as an additional parameter while keeping all others. We could include a note stating that the URL is sufficient if it exists, and that if other properties (e.g., token URL) are additionally specified, they will override the values from the well-known URL.
Do you have any further suggestions?

NGPixel:
> I actually wasn't fully aware that the discovery endpoint is optional for OIDC. For representing this in the settings, my thought would be to add the well-known URL as an additional parameter while keeping all others. We could include a note stating that the URL is sufficient if it exists, and that if other properties (e.g., token URL) are additionally specified, they will override the values from the well-known URL. @NGPixel Do you have any further suggestions?

That would work for me.

MaBauMeBad:
@H3xaChad Fist of all, thank you for your PR and the work behind it.
We are also have need for this.
Do you need any support to make the required adjustments ?

Zweisteinium:
@MaBauMeBad sorry for that long break - private stuff... :)
Also had some problems with setting up Dex to test everything
@NGPixel I did the changes as discussed: Adding the original config fields back
So the data from well-known is only fetched when specified and is overridden for any url field that has a explicit value.
When I find some time, I'll also implement the possibility to merge the id token claims for the normal Oauth2 flow as well

thomasdelorge:
@MrMeesix  : I was looking into implementing the same OIDC improvements, but you already did all the work : thanks for the PR 😁

Is there anything still left to do on this, or is it ready for review (and hopefully merge 🙏) ?
Have a good day

PR #7469, opened 2024-12-05
feat: Add support for configuring DATA_PATH from environment variable

LuisPerisVidal:


LuisPerisVidal:
@NGPixel can you accept this PR? Thanks

NGPixel:
I'm not sure I understand why you'd want to change it when running as a container? You can already map the data folder to any location outside the container. `/wiki/data/content` is already exposed as an explicit volume...

NGPixel:
Wiki.js really isn't made to run on serverless platforms... You're going to run into a lot of issues.

Also, the cache folder shouldn't be shared across instances. Using a file-share service will actually result in worst performances than fetching content from the DB directly.

PR #7493, opened 2024-12-29
Allow `HA_ACTIVE: True` to enable HA mode

Vasfed:
When something python-like (carvel ytt, ansible etc.) is used in deploying wiki - it is highly likely to have a construction similar to:
```yaml
      env:
        - name: HA_ACTIVE
          value: #@ str(data.values.replicas > 1)
```
which produces `"True"` in config and counter-intuitively does _not_ enable high-availability mode, leading to corresponding issues with caching (page edits not reflected until a lot of refreshes, randomly returning stale pages and so on - because not all workers actually refresh the page cache)

With this simple edit we can significantly decrease user frustration

### DIFF
--- a/server/core/db.js
+++ b/server/core/db.js
@@ -222,8 +222,7 @@ module.exports = {
    * Subscribe to database LISTEN / NOTIFY for multi-instances events
    */
   async subscribeToNotifications () {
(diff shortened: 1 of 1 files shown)


NGPixel (approved):

NGPixel:
I changed the logic to allow any casing for a string value.

However, if you plan on using Wiki.js in a high-availability environment, regardless of whether the replicas count is 1 or higher, HA_ACTIVE should always be set to `true`.

PR #7534, opened 2025-02-10
fix: Verify folderId when uploading files on server side

urmsm:
Add a type check for folderId when uploading files to ensure it is a number.


urmsm:
In the file upload interface (POST /u), the folderId parameter is passed into the database query without proper validation.
After being parsed by JSON.parse, it is possable not being converted to an integer. This allows malicious SQL queries to be constructed when passing parameters like `{"folderId":"id"}` or `{"folderId":{"id":0}}` (for more details, see https://www.ghostccamm.com/blog/knex_sqli/).
@NGPixel This will potentially expose Wiki.js to SQL injection risks.

PR #7625, opened 2025-05-02
Update dev image

shatovilya:
Updated development image to node 20.
Reduced image size.

### DIFF
--- a/dev/containers/Dockerfile
+++ b/dev/containers/Dockerfile
@@ -1,12 +1,13 @@
 # -- DEV DOCKERFILE --
 # -- DO NOT USE IN PRODUCTION! --
 
-FROM node:18
+FROM node:20
 LABEL maintainer "requarks.io"
 
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


NGPixel:
The version is set to 18 to make sure development is made with the oldest supported version.

BPowell76:
@NGPixel Node v18 reached EOL status on March 27, 2025. We should be using Node v20 at a minimum now if we still want Node patches.

I don't think it's a good idea to use Node v18 on a public-facing web server since any vulnerabilities won't be fixed.

NGPixel:
@BPowell76 It's not about promoting the use of v18, it's about not breaking compatibility for users that might be on v18 unless there's a really good reason to do so. Some users might be on v18 in an internal-only wiki and wouldn't want their wiki suddenly breaking because of a patch update.

BPowell76:
Given some of the support threads I have seen in the Discord, I am not sure if v18 is all that compatible though.

Issue #1496, opened 2020-02-18
Code block does not wrap text when printing


adjurk:
**Describe the bug**
Using built-in print feature, if there's a code block which contains text that is wider than the printed pdf, an unscrollable scroll bar is displayed.

**To Reproduce**
1. Create a new Wiki page
2. Paste a wide code block, for ex:

```
mkdir -p /tmp/gitea-dump/home-git-gitea-repositories && cp -r /home/git/gitea-repositories/* /tmp/gitea-dump/home-git-gitea-repositories
```

3. Press the Print button in Wiki.js -> Print to PDF
4. A pdf with an unscrollable scroll bar is generated inside that pdf.

**Expected behavior**
Text inside code block is automatically wrapped.

**Screenshots**
https://i.imgur.com/78WQ6Bp.png

**Host Info (please complete the following information):**
 - OS: Windows 10
 - Tested browsers: Google Chrome 79.0.3945.130, Microsoft Edge 44.18362.449.0 (EdgeHTML 18.18362)
 - Wiki.js version: 2.0.12

iamshalin:
do we have any update on this Bug?

mudasarellahi:
any update on this?

NGPixel:
Fixed in 2.5.308

PR #7663, opened 2025-06-04
fix Code block does not wrap text when printing

tarynlucia:


tarynlucia:
issue this fixes: https://github.com/requarks/wiki/issues/1496

NGPixel:
I moved it to `client/themes/default/scss/app.scss` in 1b6c67ee77b3ad3bf1b2cb3380c2a3eda28aa3f2 to keep it consistent with other print rules.

I also changed "word-wrap" with "overflow-wrap" which is the correct standard property to use. See [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap).

PR #7674, opened 2025-06-11
I changed the version from 18 to 20 to make it consistent with the Dockerfile in the build folder.

LuisPerisVidal:
…ckerfile in the build folder.


NGPixel:
The version is set to 18 to make sure development is made with the oldest supported version.

Issue #2324, opened 2020-08-22
Email address saved in Chrome keeps getting autofilled to top search bar


saulimus:
**Describe the bug**
The search field on top sometimes randomly autofills with my saved user/password when browsing pages in Chrome.

**To Reproduce**
1. Open site in Chrome
2. Register new user and save user/pass to Chrome
3. Go to your own user profile
4. Click around and at some point the top search bar should fill with the email address, and most likely also complain that there are no pages that fit your search query.

**Expected behavior**
The email address should only be filled when on the login page, nowhere else.

**Screenshots**
![Screenshot 2020-08-22 at 20 43 14](https://user-images.githubusercontent.com/2675404/90962386-3efb6680-e4b8-11ea-9ac4-a91d3db825a9.png)
![Screenshot 2020-08-22 at 20 42 25](https://user-images.githubusercontent.com/2675404/90962388-40c52a00-e4b8-11ea-92d4-d2b32a9034c8.png)

**Host Info (please complete the following information):**
 - OS: Mac OS X 10.14.6 + Docker
 - Wiki.js version: 2.4.107
 - Database engine: PostgreSQL 11 Alpine

**Additional context**
None.

NGPixel:
Fixed by 48d93877e33fdeb642001ade7a2a6d40d0a38aec

PR #7688, opened 2025-06-30
fix(profile): Prevent autofill on header search bar

Wingqvist:
This commit resolves the recurring issue where password managers incorrectly autofill the header search bar when viewing the user profile page.

This is achieved by wrapping the password change fields in a <form> element, properly scoping them for browser autofill logic, and using modern, standard `autocomplete` attributes. This addresses the root cause of the issue, whereas previous attempts tried to mitigate the symptom in the search bar.

Fixes [#2324](https://github.com/requarks/wiki/issues/2324)
Closes [#3327](https://github.com/requarks/wiki/discussions/3327)

### DIFF
--- a/client/components/common/nav-header.vue
+++ b/client/components/common/nav-header.vue
@@ -15,7 +15,7 @@
         prepend-inner-icon='mdi-magnify'
         :loading='searchIsLoading'
         @keyup.enter='searchEnter'
-        autocomplete='none'
+        autocomplete='off'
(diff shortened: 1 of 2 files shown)


Wingqvist:
@NGPixel Hi. I just wanted to know if I can request review from someone else, I can see you personally are busy :)

NGPixel:
Thanks!

PR #7707, opened 2025-07-17
feat: add folderByPath GraphQL resolver

dylan-hart:
Adds a new folderByPath(path: String!): AssetFolder query to the AssetQuery type in the GraphQL API. It allows clients to resolve a folder's ID and metadata by providing its hierarchical path (e.g. "uploads/images/2025"), which simplifies integration scenarios where asset folder IDs are not known in advance.

Specifically, our use case has been taking Microsoft Word documents and PDFs, converting them to Markdown, and using the GraphQL API to rapidly upload many pages and assets. Because we are inferring/generating page paths at runtime, we don't know the asset folder IDs unless we connect to the PostgreSQL database or observe network packets through the browser web dev tools.


server/graph/resolvers/asset.js
```diff
@@ -41,6 +41,25 @@ module.exports = {
...
       })
+    },
+    async folderByPath(obj, args, context) {
+      const parts = args.path.toLowerCase().split('/').filter(Boolean)
+      let parentId = null
+      for (const slug of parts) {
+        const folder = await WIKI.models.assetFolders.query().where({
+          parentId: parentId,
+          slug: slug
+        }).first()
```
NGPixel, line 52: Avoid querying the DB on every path segment. There's already a method to get the hierarchy in 1 query:
```
const folderHierarchy = await WIKI.models.assetFolders.getHierarchy(args.folderId)
```
Your code should either use it or do something similar.

NGPixel (changes requested):

server/graph/resolvers/asset.js
dylan-hart: Thanks! Would a single assetFolders.query() be acceptable? I'd like to use getHierarchy(), but it requires a folderId.

NGPixel: You're fetching all folders into memory, which isn't better. See how the getHierarchy() method works. It makes a single query using the withRecursive() method, fetching only the necessary data.

dylan-hart: Got it. Thanks!

dylan-hart: Refactored to use recursive SQL similar to getHierarchy(), starting from root and iterating downward instead of up from child/leaf. Thanks for your suggestion. Tested with PostgreSQL 15 and SQL Server 2022 (16).

PR #7714, opened 2025-07-26
upgrade mermaid from 8.8.2 to 11.9.0

JamesLavin:
Hi. Thank you for creating Wiki.js! I've been using it for years, and it has been extremely helpful!

I'm trying to use some cool features of [Mermaid.js](https://mermaid.js.org/), and only some things work. E.g., I can create simple ERDs, but defining field names & types doesn't work (see: https://mermaid.js.org/syntax/entityRelationshipDiagram.html).

I hope we can upgrade Mermaid from 8.8.2 to 11.9.0. I have attempted to do so.

I'm a backend dev who has never used `yarn` before. I *think* I succeeded, and the many changes in `yarn.lock` seem related to the Mermaid upgrade, but I noticed some packages near the end related to VS Code. I used VS Code to upgrade the `mermaid` package. I'm unsure whether packages like `vscode-languageserver-protocol@3.17.5` crept in because I used VS Code.

The only substantive change I made was running `yarn upgrade --latest mermaid`.

Thank you, again, for WikiJS!

### DIFF
--- a/package.json
+++ b/package.json
@@ -271,7 +271,7 @@
     "jest": "26.6.1",
     "js-beautify": "1.13.5",
     "js-cookie": "2.2.1",
-    "mermaid": "8.8.2",
+    "mermaid": "11.9.0",
     "mini-css-extract-plugin": "0.11.3",
     "moment-duration-format": "2.3.2",
     "moment-timezone-data-webpack-plugin": "1.3.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -106,6 +106,19 @@
     "@algolia/logger-common" "4.5.1"
     "@algolia/requester-common" "4.5.1"
 
+"@antfu/install-pkg@^1.0.0":
+  version "1.1.0"
+  resolved "https://registry.yarnpkg.com/@antfu/install-pkg/-/install-pkg-1.1.0.tgz#78fa036be1a6081b5a77a5cf59f50c7752b6ba26"
+  integrity sha512-MGQsmw10ZyI+EJo45CdSER4zEb+p31LpDAFp2Z3gkSd1yqVZGi0Ebx++YTEMonJy4oChEMLsxZ64j8FH6sSqtQ==
+  dependencies:
+    package-manager-detector "^1.3.0"
+    tinyexec "^1.0.1"
+
+"@antfu/utils@^8.1.0":
+  version "8.1.1"
+  resolved "https://registry.yarnpkg.com/@antfu/utils/-/utils-8.1.1.tgz#95b1947d292a9a2efffba2081796dcaa05ecedfb"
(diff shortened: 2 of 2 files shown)


NGPixel:
That won't work as there are breaking changes going from 8.x to 11.x.

JamesLavin:
Thank you for taking time to read and reply, @NGPixel.

[Mermaid](https://mermaid.js.org/) has indeed chosen not to maintain full backward compatibility. Some Mermaid syntax has changed, so upgrading to the latest Mermaid would break some users' existing diagrams.

But is it better to freeze Wiki.js to an outdated Mermaid version that lacks many features/improvements and can't render modern Mermaid tutorials, examples, and documentation?

Wiki.js users are capable of upgrading their existing Mermaid diagrams to use currently supported syntax. But Wiki.js users can't use features added between Mermaid 8.8.2 and 11.9.0. I've had to use Mermaid outside of Wiki.js because I need those new features. (And I was initially confused why examples I pulled straight from Mermaid tutorials were failing to render.)

Or are you saying Wiki.js itself relies on Mermaid 8.8.2 and would break if it's upgraded? If so, I'll certainly close this PR. (I have been assuming Mermaid is not used internally by Wiki.js.)

Thank you, again, for your time!

NGPixel:
That's not what I said. Mermaid 11.x requires many other components to be upgraded as well. You can't just update to 11.x and expect it to work, it won't. The implementation is different and it requires the build system to be upgraded as well. The build system can't be upgraded without breaking many other components in the process.

So it's not as simple as you think it is.

JamesLavin:
I see. I will close this PR. I apologize for not understanding the first time. I don't have lots of Javascript package management experience.

kimberlyeet:
Sorry to necro this issue, but for anyone else wondering and wanting a dirty workaround for this:

Replace this in `client/themes/default/components/page.vue`:
```
mermaid.mermaidAPI.initialize({
  startOnLoad: true,
  theme: this.$vuetify.theme.dark ? `dark` : `default`
})
```
with:
```
new Function(`return import("https://cdn.jsdelivr.net/npm/mermaid@11.14.0/dist/mermaid.esm.min.mjs")`)()
  .then(module => {
    const modernMermaid = module.default;

    modernMermaid.initialize({
      startOnLoad: false,
      theme: this.$vuetify.theme.dark ? `dark` : `default`
    });

    return modernMermaid.run({
      nodes: this.$el.querySelectorAll('.mermaid')
    });
  })
  .catch(error => {
    console.error('Failed to load modern Mermaid via CDN:', error);
  });
```

Also remove: `import mermaid from 'mermaid'`

Then build the image yourself, and voilà.

Doesn't work in the editor though and has some weird quirks.

PR #7726, opened 2025-08-04
fix: nil pointer evaluating interface {}.databaseURL

ductnn:


NGPixel:
You're introducing a condition where the object might exist but not the databaseURL, in which case nothing will work.

It would be better to keep it as a single condition and use something like [dig](https://helm.sh/docs/chart_template_guide/function_list/#dig) to perform the check.

Issue #1874, opened 2020-05-13
Sync groups and group membership through auth strategies


signalkraft:
First off: It seems there is already a few feature requests on https://wiki.js.org/feedback/?search=group for this, so I added a ticket because I'm thinking about helping with the first implementation here.

### Actual behavior
Currently authenticating with different strategies does not update group membership, even if the strategy supports that (i.e. roles in Keycloak, groups in LDAP). Manually adding users to groups is cumbersome and makes it difficult to use Wiki.js in larger teams where you want give some sub-teams their own private section.

### Expected behavior
Signing in with a strategy that supports group / role memberships should create a group on Wiki.js, if it doesn't exist yet, and then add the user to the group during authentication. There should be settings in the admin UI of the different strategies that support groups, to control this behavior. My guess would be:

* Toggle "Synchronize groups"
* Toggle "Synchronize group membership"

You could get infinitely more complex with custom group mappings, background sync of groups from LDAP, nested groups, permission mapping, etc, but as a first version the above seems useful.

--

I'd be happy to dig into the code and try to contribute a PR for LDAP and/or Keycloak, if you agree that this is a useful feature @NGPixel - it seems widely requested on Canny.

NGPixel:
Sounds good 👍

qubidt:
@signalkraft have you started on this? I'd like to give it a shot, but don't want to repeat any work you might've already done

signalkraft:
@qubidt I looked into the underlying auth lib (passport) but couldn't figure out a good way to get groups out of it. Its main purpose is authentication, so maybe I also went at this from the wrong angle.

qubidt:
@signalkraft thanks, I looked into it a bit and also came to a similar conclusion. Not sure I'm familiar enough with the codebase to make the wide changes this would require. Your solution sounds like it could work well for me so I'll give it a shot, thanks!

signalkraft:
Here's my two scripts as a starting point for anyone in a similar position. `get_ipa_client` is just a thin wrapper around FreeIPAs Python SDK.

```python

import os

import requests


def run_query(query, variables=None):
    headers = {"Authorization": f"Bearer {os.environ.get('WIKIJS_API_KEY')}"}
    json = {'query': query}
    if variables:
        json['variables'] = variables
    request = requests.post(os.environ.get('WIKIJS_API'), json=json, headers=headers)
    if request.status_code == 200:
        return request.json()
    else:
[...]

PR #7736, opened 2025-08-11
feat: Set groups based on Entra groups (Azure auth module)

pianosaurus:
# Add group synchronization for Entra ID (aka. Azure AD)

Azure AD is now called Entra, hence the name of the auth module.

This is a copy-paste of the code used from generic OAuth2, except the group claim name is always "groups". This is not configurable in Entra, so no configuration is needed here either. It solves #1874 for the Azure module, as others have already done for SAML and OIDC.

# User how-to

1. Enable the "groups" claim in Entra under App Registrations ("Token configuration" -> "Add groups claim").
2. Edit the Azure auth module in Wiki.js, and enable the new setting ("Map groups").
3. Create the groups in Wiki.js.
4. Groups are mapped during the auth process, so users have to log in again.

### DIFF
--- a/server/modules/authentication/azure/authentication.js
+++ b/server/modules/authentication/azure/authentication.js
@@ -48,6 +48,19 @@ module.exports = {
               picture: ''
             }
           })
+          if (conf.mapGroups) {
+            const groups = _.get(profile, '_json.groups')
+            if (groups && _.isArray(groups)) {
+              const currentGroups = (await user.$relatedQuery('groups').select('groups.id')).map(g => g.id)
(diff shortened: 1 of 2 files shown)


NGPixel (approved):

NGPixel:
Thanks!

elypse-dev:
Thank a lot for this improvement merged a few weeks before I start testing wiki.js
Therefore, I had to do a few things before it work as I hoped
Entra seems to send group names only in case of synced groups. In case of pure cloud groups, Entra sends the ID
Moreover, I find these name/id in the roles section, not in the group section
As a test, I add the id in the wiki.js group name between brackets and replace your code by this one :
image
And I works as I expected

PR #7745, opened 2025-08-14
Switch service port on the exposed container port

ThomasSamson:
Hi, 

Just a little pull request for the helm chart. 
Switching the service to match with the container exposed port.


NGPixel:
The service port doesn't need to match. It's in fact very common to expose a service at port 80.

PR #7784, opened 2025-09-11
feat(chart): add custom environment variables to the deployment

lyz-code:
Hi, first of all thanks for creating and maintaining both wikijs and it's helm chart.

For our deployment we need to use the `PGSSLMODE:   require` environment variable in the deployment. Instead of adding that option I've found more potentially useful for other users to let thme define environment variables.

Thanks!


NGPixel:
Thanks!

PR #7792, opened 2025-09-22
fix: rocket.chat auth state and endpoint

sanyappc:


y0uCeF:
I tested this PR and Rocket.Chat login is working for me on my wiki.js installation. I even got my profile picture on my wiki.js account and was assigned to my default Group.
Thanks for the quick fix

PR #7973, opened 2026-04-02
fix: multiple vega branch fixes for ESM modules, auth, and admin UI

gmowses:
## Summary

Bug fixes found while deploying Wiki.js 3.0 (vega branch) in production on AWS EKS.

### ESM/CJS Module Compatibility
- Auth and storage modules use `require()` (CJS) but `package.json` has `"type": "module"`. Node.js 24 treats all `.js` files as ESM, causing `require is not defined` errors. Fixed via `.cjs` rename + ESM wrapper in Dockerfile.

### Auth Controller
- `/login/:strategy` route passes `{ strategy }` to `users.login()` but the method expects `{ strategyId }`. This causes `AuthProviderInvalid` for all OAuth/OIDC logins.

### WIKI.Error Undefined
- `WIKI.Error.AuthLoginFailed()` and `WIKI.Error.AuthProviderInvalid()` crash with `Cannot read properties of undefined` because `WIKI.Error` is not initialized in the vega branch. Replaced with standard `new Error()`.

### Null-safe .trim()
- `updatePage` crashes with `Cannot read properties of null (reading 'trim')` when `title`, `description`, `icon`, or `alias` is null in the patch. Added `?? ''` before `.trim()`.

### processProfile for v3 Schema
- `processProfile` queries `providerId` and `providerKey` columns which don't exist in the v3 schema. Updated to use email-based user lookup.

### OIDC Verify Callback Signature
- `passport-openidconnect` calls the verify callback with 10 arguments (req, iss, uiProfile, idProfile, context, idToken, accessToken, refreshToken, params, cb) but the OIDC module only accepts 4 (req, iss, sub, profile, cb), causing `accessToken` to be undefined.

### OAuth Error Handling
- OAuth callback errors cause `Internal Server Error` (500) because `next(err)` tries to render a template but no view engine is configured. Changed to redirect to `/login?error=message`.

### Admin UI - GroupEditOverlay
- Save button has no `@click` handler (empty)
- `assignUser()` function is empty
- `unassignUser()` function is empty

## Test Plan
- [x] Deployed and tested on AWS EKS with PostgreSQL Aurora
- [x] OIDC login with self-hosted GitLab
- [x] Page create/edit/save
- [x] Admin group management


gmowses:
Closing - PR contains fork customizations mixed with bug fixes. Will reopen with clean bug-fix-only commits.

PR #7974, opened 2026-04-02
fix: multiple vega branch bug fixes for auth, pages, mail

gmowses:
## Summary

Bug fixes found while deploying Wiki.js 3.0 (vega branch) in production on AWS EKS with Node.js 24.

**No UI changes, no customizations - server-side fixes only.**

### 1. Auth Controller - strategyId field name (`server/controllers/auth.mjs`)
`/login/:strategy` and `/login/:strategy/callback` routes pass `{ strategy: req.params.strategy }` to `users.login()`, but the method destructures `{ strategyId }`. This causes all OAuth/OIDC logins to fail with `AuthProviderInvalid`.

### 2. Auth Controller - Error handling (`server/controllers/auth.mjs`)
OAuth callback errors cause `Internal Server Error` (500) because `next(err)` tries to render an error template but no view engine is configured in the vega branch. Changed to redirect to `/login?error=message`.

### 3. WIKI.Error undefined (`server/models/users.mjs`)
`WIKI.Error.AuthLoginFailed()`, `WIKI.Error.AuthProviderInvalid()`, `WIKI.Error.AuthAccountBanned()` all crash with `Cannot read properties of undefined (reading 'AuthLoginFailed')` because `WIKI.Error` is not initialized. Replaced with standard `new Error()` calls.

### 4. Null-safe .trim() (`server/models/pages.mjs`)
`updatePage` crashes with `Cannot read properties of null (reading 'trim')` when `title`, `description`, `icon`, or `alias` is null in the update patch. Added `?? ''` before `.trim()`.

### 5. Mail template rendering (`server/core/mail.mjs`)
`@vue-email/compiler` `render()` returns `{ html: string, text: string }` but `loadTemplate()` returns the object directly to nodemailer's `html` field, causing `The "chunk" argument must be of type string`. Now extracts `result.html`.

## Test Environment
- AWS EKS (Kubernetes 1.29)
- Node.js 24
- PostgreSQL 16.8 (Aurora)
- OIDC with self-hosted GitLab

### DIFF
--- a/server/controllers/auth.mjs
+++ b/server/controllers/auth.mjs
@@ -42,10 +42,11 @@ export default function () {
   router.get('/login/:strategy', async (req, res, next) => {
     try {
       await WIKI.db.users.login({
-        strategy: req.params.strategy
+        strategyId: req.params.strategy
       }, { req, res })
     } catch (err) {
-      next(err)
+      WIKI.logger.error(`OAuth login error: ${err.message}`)
+      res.redirect('/login?error=' + encodeURIComponent(err.message))
     }
   })
 
@@ -57,7 +58,7 @@ export default function () {
 
     try {
       const authResult = await WIKI.db.users.login({
-        strategy: req.params.strategy
+        strategyId: req.params.strategy
       }, { req, res })
       res.cookie('jwt', authResult.jwt, { expires: DateTime.now().plus({ years: 1 }).toJSDate() })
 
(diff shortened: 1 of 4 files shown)


NGPixel:
As stated in the vega branch README, it's buggy, incomplete and not secure. **You should NOT be running it in production**. Bugs should not be reported for this branch, as it's not helpful in it's current state.

PR #8005, opened 2026-05-12
fix: null-guard storage targets and page tags to prevent crashes

derekderie:
## Problem

Two places in the storage layer can throw a `TypeError: Cannot read properties of undefined` when iterating over collections that may not be initialized:

  - `Storage.targets` can be `undefined` when storage hasn't been
    activated/configured yet. Affects `pageEvent`, `assetEvent`, and
    `getLocalLocations`.
  - `currentPage.tags` can be `undefined` when a page was imported or
    created without tags. Affects the disk storage `common.js` handler.

## Fix

Add `|| []` fallback guards so these methods degrade gracefully instead of throwing:

  - `server/models/storage.js` — 3 occurrences
  - `server/modules/storage/disk/common.js` — 1 occurrence

### DIFF
--- a/server/models/storage.js
+++ b/server/models/storage.js
@@ -179,7 +179,7 @@ module.exports = class Storage extends Model {
 
   static async pageEvent({ event, page }) {
     try {
-      for (let target of this.targets) {
+      for (let target of (this.targets || [])) {
         await target.fn[event](page)
       }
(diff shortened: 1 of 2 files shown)


derekderie:
Found the root cause of #7978

PR #8006, opened 2026-05-12
fix: include full stack traces in error logs

derekderie:
## Problem

Error logs across all job handlers only logged `err.message`, which drops the stack trace entirely. This makes it very hard to diagnose where errors originate.

## Fix

  - `server/core/logger.js`: add `winston.format.errors({ stack: true })`
    to the format chain, and update the printf format to append the stack
    when present.
  - `server/jobs/*.js` (8 files): pass the full `err` object to
    `WIKI.logger.error()` instead of `err.message`.

### DIFF
--- a/server/core/logger.js
+++ b/server/core/logger.js
@@ -7,6 +7,7 @@ module.exports = {
   loggers: {},
   init(uid) {
     const loggerFormats = [
+      winston.format.errors({ stack: true }),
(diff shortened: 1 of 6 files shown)


derekderie:
Experienced a lot of trouble in debugging #7978, had to create this PR in order to successfully uncover the root cause of the wiki crashes.

PR #8008, opened 2026-05-12
chore: fix CVE-2021-3766 by upgrading to objection v3

LaurentGoderre:


NGPixel:
This is a major breaking change and wouldn't work.

PR #8035, opened 2026-06-22
Add Gateway API HTTPRoute support to the Wiki.js Helm chart

somaz94:
This adds an optional Gateway API HTTPRoute template to the Wiki.js Helm
chart, as an alternative to the existing Ingress for clusters that route
with a Gateway controller. It is disabled by default (httpRoute.enabled:
false), so clusters without the Gateway API CRDs are unaffected.

Changes:
- New templates/httproute.yaml, gated on httpRoute.enabled, backed by the
  existing Service (service.port).
- httpRoute block in values.yaml (enabled, annotations, hostnames,
  parentRefs) with inline comments.
- README: parameter table rows plus an "HTTPRoute (Gateway API)" section.

Validation (helm v3.18.1):
- helm lint --strict: pass
- httpRoute.enabled=false (default): nothing rendered
- httpRoute.enabled=true with parentRefs and hostnames: valid HTTPRoute
- enabled with hostnames omitted: hostnames field correctly dropped
- full chart render: no regressions

### DIFF
--- a/dev/helm/README.md
+++ b/dev/helm/README.md
@@ -112,6 +112,10 @@ The following table lists the configurable parameters of the Wiki.js chart and t
 | `ingress.annotations`                | Ingress annotations                         | `{}`                                                       |
 | `ingress.hosts`                      | List of ingress rules                        | `[{"host": "wiki.local", "paths": ["/"]}]`                |
 | `ingress.tls`                        | Ingress TLS configuration                   | `[]`                                                       |
+| `httpRoute.enabled`                  | Enable Gateway API HTTPRoute resource       | `false`                                                    |
+| `httpRoute.annotations`              | HTTPRoute annotations                       | `{}`                                                       |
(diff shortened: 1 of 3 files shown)


somaz94:
Gentle nudge on this one — it's been a few weeks. This adds opt-in Gateway API HTTPRoute support to the Helm chart (default off → existing installs are unchanged). Happy to adjust anything if it helps. Thanks for maintaining Wiki.js!

somaz94:
Closing this for now to keep the queue tidy — happy to reopen if there's interest in Gateway API HTTPRoute support for the Wiki.js chart later.

PainOchoco:
I'm interested in this :)
Ingress NGINX is deprecated since March 2026 (https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/), and it would be great to have Gateway API support in the helm chart.
Thanks!

PR #8061, opened 2026-08-05
feat(storage): add Google Cloud Storage module

horusglez:
## Summary

Adds a push-mode storage target module for Google Cloud Storage. Wiki.js core currently ships storage targets for Azure, S3, Dropbox, Box, Google Drive, OneDrive, DigitalOcean Spaces, SFTP, Git and local disk, but not GCS directly.

Follows the same interface as the existing `disk`/`s3` modules: `created`, `updated`, `deleted`, `renamed`, `assetUploaded`, `assetDeleted`, `assetRenamed`, plus an `exportAll` action (mirrors `disk`'s `dump`) to bulk-push existing content.

## Config

- Bucket Name
- Project ID
- Service Account Key (JSON)
- Make Uploaded Files Public (optional)

The bucket must already exist; the module validates access via `bucket.exists()` on init but does not create the bucket.

**Required service account roles (on the bucket, not project-wide):** `Storage Object Admin` (object CRUD) + `Storage Legacy Bucket Reader` (needed for `storage.buckets.get`, used by the init check — `Storage Object Admin` alone does not include this permission).

## Test plan

- [x] Unit tests added in `server/test/modules/storage/gcs/storage.test.js` covering `init` (valid/invalid JSON key, missing/inaccessible bucket, projectId fallback to the key's `project_id`), file path resolution including locale namespacing, and all create/update/delete/rename/asset handlers, with `@google-cloud/storage` mocked.
- [x] Manually validated against a real GCS bucket: module loads (`Loaded 1 new storage targets: OK`), init succeeds against a real bucket with the two roles above, page create/update/delete correctly reflected in the bucket, confirmed visually in the GCS console.

🤖 Built with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "filesize": "6.1.0",
     "fs-extra": "9.0.1",
     "getos": "3.2.1",
+    "@google-cloud/storage": "7.21.0",
     "graphql": "15.3.0",
     "graphql-list-fields": "2.0.2",
     "graphql-rate-limit-directive": "1.2.1",
--- a/server/modules/storage/gcs/definition.yml
+++ b/server/modules/storage/gcs/definition.yml
@@ -1,0 +1,42 @@
+key: gcs
+title: Google Cloud Storage
+description: Google Cloud Storage is a fully-managed, scalable object storage service offered by Google Cloud Platform.
+author: Horus Gonzalez (built with Claude)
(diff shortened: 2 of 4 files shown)


NGPixel:
No new features are accepted for the v2 branch.

GCS was added in v3 already.

PR #8085, opened 2026-09-05
Add collapsible spoiler blocks to Markdown

Mhfd10:
Adds `::: spoiler` blocks to the Markdown editor preview and server renderer, using native HTML `<details>` and `<summary>` elements. When editing a wiki page this is more intuitive to me than using html.

Example:

```md
::: spoiler Reveal answer
Hidden **Markdown content**.
:::
```

The title defaults to “Spoiler” when omitted. Custom titles are HTML-escaped.
Adds `markdown-it-container@3.0.0` and its lockfile entry, plus tests for default titles, escaped custom titles, and unrelated container syntax.

### DIFF
--- a/client/components/editor/editor-markdown.vue
+++ b/client/components/editor/editor-markdown.vue
@@ -224,6 +224,7 @@ import mermaid from 'mermaid'
 
 // Helpers
 import katexHelper from './common/katex'
+import spoiler from './markdown/spoiler'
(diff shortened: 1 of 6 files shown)


NGPixel:
This feature has already been added in v3. It will not be backported to v2 as it introduces a new MDC concept.

PR #8091, opened 2026-09-15
fix: only reject locale-like asset folder names at the root

djlongy:
The new asset folder dialog refuses any two-letter name. `isFolderNameValid` tests it with `localeSegmentRegex`, the same test `parsePath` uses to shift a locale off the front of a URL, but `parsePath` only inspects the first segment, so a nested folder is safe. In discussion #4038 the answer was that the restriction only applies to the root folder. The code applies it at every depth, because the computed never reads the parent.

This gates the locale test on `currentFolderId > 0`, the idiom the file already uses at line 376. Root names are unchanged: `ui`, `en` and `en-us` are still refused there, where the collision is real. `parsePath('/ui/x.png')` returns locale `ui` and path `x.png`, the 404 reported on #1292.

Verified by evaluating the computed at both depths and running `parsePath` directly. eslint reports no new warnings. No test: jest here is server-only and CI runs only the Cypress installer spec.

Refs #1292


djlongy:
Evidence, since this was reported once before and closed.

### Steps to reproduce on 2.5.x

Open any page in the editor, click Assets, then New Folder. Type `ui`. Create stays disabled. Navigate into any existing folder and try `ui` again: still disabled. With this patch the second case is enabled and the first is not.

### The computed, evaluated at both depths

```
name      root:before  root:after   sub:before  sub:after
ui        false        false        false       true
ux        false        false        false       true
db        false        false        false       true
en        false        false        false       true
en-us     false        false        false       true
uix       true         true         true        true
assets    true         true         true        true
u         false        false        false       false
(empty)   false        false        false       false
```

Nothing at the root changes. `en-us` being refused at five characters is why this is not a length rule, though the docs describe it as one ("Must be at least 2 characters in length").

### Why the root case must keep the check

Running `parsePath` from `server/helpers/page.js` unmodified, with `WIKI.config.lang.code = 'en'`:

```
/ui/ui-assets-dialog.png         -> locale=ui     path=ui-assets-dialog.png
/assets/ui/ui-assets-dialog.png  -> locale=en     path=assets/ui/ui-assets-dialog.png
/uix/ui-assets-dialog.png        -> locale=en     path=uix/ui-assets-dialog.png
/db/x.png                        -> locale=db     path=x.png
/en-us/x.png                     -> locale=en-us  path=x.png
/u/x.png                         -> locale=en     path=x.png
```

A root folder named `ui` loses its first segment, so `assets.getAsset` hashes the wrong path and the file 404s. That is what juliendms reported on #1292 in March 2020. Nested under `assets` the path survives intact, which is the case #4038 asked about.

`server/controllers/common.js:419` calls `parsePath` before deciding page or asset, so both go through the same two shifts, and both only look at segment 0. The dialog's two rules mirror those two shifts exactly: `length > 1` for the single-character shift, the regex for the locale shift.

### The documentation is published under this layout

```
$ curl -sI https://docs.requarks.io/assets/ui/ui-assets-dialog.png
200  image/png  136831 bytes
```

### 3.x resolved it by asking a different question

`splitLocalePath` in the 3.x backend tests a segment against the site's registered locale prefixes rather than against a shape:

```ts
const locale = prefixes.get(first)
if (!locale) { return null }
```

On 3.0.0-beta.561 a folder named `ui` creates, an upload to `folderPath=ui` serves back byte-identical, and a page at `db/postgres` stores as `db/postgres` rather than locale `db` plus page `postgres`. The same change closes the page-side half of #1292. Backporting that to 2.x would mean touching `parsePath`, which is a routing change, so this patch stays in the dialog.

### Not touched

`@keyup.enter='createFolder'` on line 28 bypasses `isFolderNameValid`, so Enter already creates a folder the button refuses. Pre-existing and out of scope here.

djlongy:
Reproduced on a running 2.5.314 (`requarks/wiki:2.5.314` with postgres:15), rather than from the source alone.

### The server accepts what the dialog refuses

Signed in as admin, raw GraphQL to `/graphql`:

```
mutation { assets { createFolder(parentFolderId: 0, slug: "ui") { responseResult { succeeded message } } } }
-> { "succeeded": true, "message": "Asset Folder has been created successfully." }
```

Same for `db`, and for `ui` under `assets` (parentFolderId 2). `assets.folders` then lists them. So the restriction exists only on the button.

### The root collision, measured

A 69-byte PNG uploaded into each folder, then fetched at the URL the editor inserts (`editor-modal-media.vue:373-376`):

```
/ui/x.png         404 text/plain     (root folder "ui")
/db/x.png         404 text/plain     (root folder "db")
/uix/x.png        200 image/png      (control)
/assets/ui/x.png  200 image/png      ("ui" nested)
/en/ui/x.png      200 image/png      (only one segment is ever shifted)
```

This install had `namespacing: false` and no namespaces configured, and `db` is not a registered locale, yet `/db/x.png` still 404s. The shift in `parsePath` is shape-based, not a lookup against configured locales. That is why this patch gates on depth rather than testing the name against a locale list: a list would let a user create root `db`, whose assets would then 404.

### The dialog, in a browser

Inside `/assets/` (`currentFolderId` = 2), typing `ui` into Folder Name gives `isFolderNameValid: false` and Create stays `disabled`. That is the subfolder case #4038 describes as allowed.

With the one-line expression from this PR substituted into the computed at runtime (the shipped bundle in the image cannot be rebuilt, so this is a runtime substitution and not a rebuilt client): same folder, same typed name, `isFolderNameValid: true` and Create enabled. Navigating back up to the root with `ui` still typed returns `isFolderNameValid: false`, Create disabled.

Refused on main, accepted with the change, still refused at the root.

